PR #3, opened 2016-09-29
Fix/show data

seromenho:


js/interface.js
```diff
@@ -54,18 +54,18 @@ function fetchCurrentDataSourceEntries() {
...
 
-    var tableHead = '<tr>' + Object.keys(rows[0]).map(function (column) {
+    var tableHead = '<tr>' + Object.keys(rows[0].data).map(function (column) {
       return '<td>' + column + '</td>';
     }).join('') + '</tr>';
 
     var tableBody = rows.map(function (row) {
-      return '<tr>' + Object.keys(row).map(function (key) {
-        return '<td>' + row[key] + '</td>';
+      return '<tr>' + Object.keys(row.data).map(function (key) {
```
squallstar: why? this will break when you save data.

js/interface.js
```diff
@@ -53,18 +54,18 @@ function fetchCurrentDataSourceEntries() {
...
+        return '<td>' + row.data[key] + '</td>';
```
squallstar: I'll also add something like `<td>' + typeof row[key] === 'object' ? JSON.stringify(row[key]) : row[key] + '</td>'`

js/interface.js
seromenho: Because getting data from a datasource return the da source entry raw, not only the data.
This is good as we might need updatedAt/createdAt in a very near future.

js/interface.js
seromenho: Or a recursive table 😁.
I'll prefer your option for now.

js/interface.js
squallstar: I understand but then how come it was working before? I thought it should then have displayed weird rows instead of data.

js/interface.js
squallstar: Recursive table is definitely the way to go! Let's just put it as a ticket here on GitHub on this repo, we'll figure it out soon.

js/interface.js
seromenho: It worked for some time. It wasn't working for a while due to that change on the API. Shame on me I didn't fix this at the time of the API change. Shame shame
Sorry :)

seromenho:
I have made the requested changes.

PR #11, opened 2017-01-25
Updates to interface

tonytlwu:
- Adds “Save changes” button
- Makes table-entries area scrollable
- Adds floating toolbar


tonytlwu:
This tackles points 1, 3 and 5 at https://github.com/Fliplet/fliplet-studio/issues/636

There's still a long delay when loading large datasets. We should get a replacement for TinyMCE

PR #16, opened 2017-03-21
Don't show menus datasources

seromenho:


js/interface.js
```diff
@@ -58,7 +59,16 @@ function getDataSources() {
   $dataSources = $('#data-sources > tbody');
 
   Fliplet.DataSources.get().then(function (dataSources) {
```
squallstar: didn't we have a `type` filter option?

seromenho: :trollface:

squallstar: just checked, we do. https://github.com/Fliplet/fliplet-api/blob/master/routes/v1/data-sources.js#L78

seromenho: The filter is to choose only menus.
But we can enhance it and check if null then send only null types.

squallstar: `get({ type: 'menu' })`?

seromenho: That works.
But here you want to get `type: null` because "data" data sources have no type.

squallstar: we pass it to sequelize so you can do `{ type: { $notIn: ['menu'] } }`

squallstar: yeah or `type: null`

seromenho:
Fliplet/fliplet-widget-data-source/pull/17

PR #18, opened 2017-04-10
Don't cache data sources

seromenho:
When changing between data sources list should be refreshed from API
Fliplet/fliplet-studio#711


seromenho:
Done in Fliplet/fliplet-widget-data-source#20

PR #19, opened 2017-04-10
Don't render again list of data sources

seromenho:
If they have been rendered once.
Fliplet/fliplet-studio#711


seromenho:
Fliplet/fliplet-widget-data-source/pull/20

PR #28, opened 2017-11-03
Feat handsontable

seromenho:


seromenho:
Continued work here: https://github.com/Fliplet/fliplet-widget-data-source/pull/29

PR #32, opened 2018-01-18
Fixes #2036

hcarneiro:
we forgot to remove the header from the interface, as it became part of Studio and refactor the resize function.

The user couldn't see the end of the table.

Fixes https://github.com/Fliplet/fliplet-studio/issues/2036


squallstar:
This also fixes https://github.com/Fliplet/fliplet-studio/issues/2021

PR #37, opened 2018-01-31
Chore/enhance find

seromenho:
- Type to highlight
- `Cmd/Ctrl + F` to focus on search field
- `Enter` or `Cmd/Ctrl + G` to find next result
- `Shift + Enter` or `Cmd/Ctrl + Shift + G` to find previous result
- Wraps result when user reaches the first/last result

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -361,12 +361,19 @@ var resultsCount = 0;
  * This will make a search
  * @param {string} type next | prev | find | clear
  */
+var previousSearchValue = '';
 function search(type) {
   if (type === 'clear') {
     searchField.value = '';
   }
 
   var value = searchField.value;
+  //  Don't run search again if the value hasn't changed
+  if (type === 'find' && previousSearchValue === value) {
+    return;
+  }
+  previousSearchValue = value;
+
   if (value !== '') {
     $('.filter-form .find-controls').removeClass('disabled');
   } else {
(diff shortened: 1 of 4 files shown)


js/spreadsheet.js
```diff
@@ -351,14 +351,101 @@ var spreadsheet = function(options) {
...
-var searchFiled = document.getElementById('search-field');
-Handsontable.dom.addEvent(searchFiled, 'keyup', function (event) {
-  var queryResult = hot.search.query(this.value);
-  var resultsCount = queryResult.length;
-  $('.find-results').html(resultsCount + ' found');
-  hot.render();
+var searchField = document.getElementById('search-field');
+
+// Prevent cmd+f default behaviour and use our find
+window.addEventListener("keydown", function (event) {
```
tonytlwu: @seromenho Use `keyup`

js/spreadsheet.js
```diff
@@ -351,14 +351,101 @@ var spreadsheet = function(options) {
...
+  var ctrlDown = (event.ctrlKey || event.metaKey) && !event.altKey;
+
+  if (!ctrlDown) {
+    return;
+  }
+
+  if (event.keyCode === 114 || (ctrlDown && event.keyCode === 70)) { 
+    event.preventDefault();
+    hot.deselectCell();
```
tonytlwu: I think this'll trigger an error if the user is looking at the data source list, because HOT isn't initialised

js/spreadsheet.js
tonytlwu: Ignore that. Use `keydown` because we need to capture the modifier.

js/spreadsheet.js
tonytlwu: `TypeError: Cannot read property 'deselectCell' of undefined`

js/spreadsheet.js
```diff
@@ -351,12 +351,123 @@ var spreadsheet = function(options) {
...
+  if (resultsCount) {
+    foundMessage = (queryResultIndex + 1) + ' of ' + foundMessage;
+  }
+  $('.find-results').html(value !== '' ? foundMessage : '');
+  
+  // Focus back to the search field
+  searchField.focus();
+}
+
+search.bind(searchField);
```
seromenho: we don't need this anymore

PR #39, opened 2018-04-18
Removes duplicated apps and adds search by ID

tonytlwu:


js/interface.js
```diff
@@ -178,9 +184,13 @@ function saveCurrentData() {
 }
 
 // Append a data source to the DOM
-function renderDataSource(data) {
+function renderDataSource(data, options) {
+  options = options || {};
   var tpl = Fliplet.Widget.Templates['templates.dataSource'];
   var html = tpl(data);
```
squallstar: you could simply edit `data.apps` here to avoid duplication and all these code changes, no?

tonytlwu: Can do 🙆‍♂️

tonytlwu: @squallstar Dried

PR #41, opened 2018-07-13
Fixes issue with jumping column width

hcarneiro:
Also tested opening after saving with huge rows. Table keeps its share and when editing the cell it focus and scroll to the end. No more jumping around.

screen recording 2018-07-13 at 05 58 pm


tonytlwu:
@hcarneiro Are column widths still manually adjustable after this or are they locked to 250px?

hcarneiro:
@tonytlwu still manually adjustable without restrictions.

screen recording 2018-07-13 at 06 08 pm

PR #45, opened 2018-10-31
Allow the data sources widget to initialise filtered by appId

hcarneiro:
This enhancement is for when a user clicks on App Data from the right side bar in Edit, the overlay with the list of data sources, only shows the data sources used by the app.

Studio commit reference: https://github.com/Fliplet/fliplet-studio/pull/3386/commits/f5eb3d421e555ff0ee7d3e0e0efe95b16c008769

screenshot 2018-10-31 at 13 19 53

### DIFF
--- a/css/interface.css
+++ b/css/interface.css
@@ -32,10 +32,26 @@ body {
   border: none;
 }
 
+#data-sources th.desc,
+#data-sources th.asc {
+  cursor: pointer;
+}
+
+#data-sources th.desc .fa-sort-amount-asc,
+#data-sources th.asc .fa-sort-amount-desc {
+  display: none;
+}
+
+#data-sources th.desc .fa-sort-amount-desc,
+#data-sources th.asc .fa-sort-amount-asc {
+  display: inline-block;
+}
+
 #data-sources td {
   border-left: none;
   border-right: none;
   border-bottom: none;
+  vertical-align: middle;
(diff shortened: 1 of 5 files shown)


ibroom:
@hcarneiro will you be including a way that users can select an app in the data source manager to see only data sources related to that app? e.g. an interactive way to apply the filter?

hcarneiro:
@tonytlwu This last commit adds the following:
- Added Last updated column
- Allow users to sort by Last updated or Data source name
- Added a button to see all the data sources when see the list filtered by the app
- UI tweaks (vertically align rows, remove the `<code>` from the IDs)

hcarneiro:
@ibroom We just did some other enhancements but we didn't think of that scenario. Can be done though!

tonytlwu:
@hcarneiro I'd say **a way that users can select an app in the data source manager to see only data sources related to that app** is actually the more useful feature compared to the other changes. If @ibroom wasn't clear, I think it should've been the only thing that needed implementing. Can we implement it as part of this PR?

hcarneiro:
@tonytlwu As with anything we could add as many features as we can think of, but this PR isn't for that. This will already filter the data sources in context of the app you are editing. We can later enhance the component so that when accessed by the App Data link found in the apps list page, the user can filter by app.

PR #47, opened 2019-02-11
Logs and renders errors when loading data sources

tonytlwu:
Ref. https://github.com/Fliplet/fliplet-e2e-tests/issues/272

Initial error message:

image

If user clicks **Details**...

image

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -38,12 +38,12 @@ function getDataSources() {
     return;
   }*/
 
-  Fliplet.DataSources.get({
-      roles: 'publisher,editor',
-      type: null
-    }, {
-      cache: false
-    })
+  return Fliplet.DataSources.get({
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -82,11 +99,51 @@ function renderDataSources(dataSources) {
...
+    Fliplet.Modal.alert({
+      message: options.message
+    });
+    return;
+  }
+
+  Fliplet.Modal.confirm({
+    message: options.message,
+    buttons: {
+      cancel: {
```
squallstar: does pressing "Details" actually do anything?

squallstar: ah, nevermind. I've seen what you're done. It's a bit hard to read but works.

tonytlwu: Yes, it will show the parsed error message (if it exists).

PR #52, opened 2019-10-21
Now user cannot create data source with empty name (spaces)

MaksymShokin:
@tonytlwu @squallstar 

## Issue
Fliplet/fliplet-studio#3773

## Description
Changed confirm to Fliplet.Modal.confirm. Added alert when the user is trying to submit empty (only spaces) name.

## Screenshots/screencasts
edit empty ds

## Backward compatibility

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -560,21 +560,26 @@ $('#app')
   })
   .on('click', '[data-create-source]', function(event) {
     event.preventDefault();
-    var sourceName = prompt('Please type the new table name:');
+    Fliplet.Modal.prompt({
+      title: 'New data source',
+      message: 'Enter a data source name',
+    }).then(function (result) {
+      var dataSourceName = result.trim();
+      if (!dataSourceName) {
+        Fliplet.Modal.alert({ message: 'You must enter a data source name'});
+        return;
+      }
 
-    if (!sourceName) {
-      return;
-    }
-
-    Fliplet.Organizations.get().then(function(organizations) {
-      return Fliplet.DataSources.create({
-        organizationId: organizations[0].id,
-        name: sourceName
+      Fliplet.Organizations.get().then(function(organizations) {
+        return Fliplet.DataSources.create({
+          organizationId: organizations[0].id,
+          name: dataSourceName
+        });
+      }).then(function(createdDataSource) {
+        dataSources.push(createdDataSource);
+        $dataSources.append(getDataSourceRender(createdDataSource));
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -560,21 +560,26 @@ $('#app')
...
-
-    if (!sourceName) {
-      return;
-    }
+    Fliplet.Modal.prompt({
+      title: 'New data source',
+      message: 'Enter a data source name',
+    }).then(function (result) {
+      var dataSourceName = result.trim();
+      if (!dataSourceName) {
```
tonytlwu: The result could come back with `null` (cancelled) or `""`. They should be treated differently.

If it's cancelled, no further action should be taken, including any alerts.

If it comes back with an empty string, the alert defined here should be shown. After the alert is dismissed, the user should be shown the same prompt modal again so they can continue.

tonytlwu: The result could come back with `null` (cancelled) or `""`. They should be treated differently.

If it's cancelled, no further action should be taken, including any alerts.

If it comes back with an empty string, the alert defined here should be shown. After the alert is dismissed, the user should be shown the same prompt modal again so they can continue.

MaksymShokin: Fixed

js/interface.js
```diff
@@ -360,6 +360,37 @@ function browseDataSource(id) {
     });
 }
 
+function createDataSource() {
+  Fliplet.Modal.prompt({
+  title: 'New data source',
```
squallstar: indentation is incorrect

MaksymShokin: Fixed

squallstar:
@tonytlwu bump ^

js/interface.js
```diff
@@ -360,6 +360,37 @@ function browseDataSource(id) {
...
 
+function createDataSource() {
+  Fliplet.Modal.prompt({
+    title: 'New data source',
+    message: 'Enter a data source name',
+  }).then(function (result) {
+    if (result === null) {
+      return;
+    }
+    var dataSourceName = result.trim();
```
tonytlwu:
```suggestion

    var dataSourceName = result.trim();
```

js/interface.js
```diff
@@ -360,6 +360,37 @@ function browseDataSource(id) {
...
+    if (!dataSourceName) {
```
tonytlwu:
```suggestion

    if (!dataSourceName) {
```

PR #58, opened 2019-10-28
Before deleting data source modal window will show apps where this data source is used

MaksymShokin:
## Issue
Fliplet/fliplet-studio#3774

## Description
Replaced default confirm with Fliplet.confirm. Added names of apps where the data source is used.

## Screenshots/screencasts
deleting demo

## Backward compatibility

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -536,9 +536,27 @@ $('#app')
   })
   .on('click', '[data-delete-source]', function(event) {
     event.preventDefault();
-    var confirmAlert = confirm('Are you sure you want to delete this data source? All entries will be deleted.');
+    var message = 'Are you sure you want to delete this data source? All entries will be deleted.';
+    var currentDS = _.find(dataSources, function(ds) {
+      return ds.id === currentDataSourceId;
+    });
 
-    if (confirmAlert) {
+    if (currentDS.apps.length) {
+      var appPrefix = currentDS.apps.length > 1 ? 'apps: ' : 'app: ';
+      var appUsedIn = currentDS.apps.map(function(elem) {
+        return ' ' + elem.name;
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -536,9 +536,27 @@ $('#app')
...
-    var confirmAlert = confirm('Are you sure you want to delete this data source? All entries will be deleted.');
+    var message = 'Are you sure you want to delete this data source? All entries will be deleted.';
+    var currentDS = _.find(dataSources, function(ds) {
+      return ds.id === currentDataSourceId;
+    });
+
+    if (currentDS.apps.length) {
+      var appPrefix = currentDS.apps.length > 1 ? 'apps: ' : 'app: ';
+      var appUsedIn = currentDS.apps.map(function(elem) {
+        return ' ' + elem.name;
```
squallstar: just return `elem.name` and then join the results with `.join(', ')`

js/interface.js
```diff
@@ -536,9 +536,27 @@ $('#app')
...
+      });
+
+      message = 'Are you sure you want to delete this data source? The data source is used in the next ' + appPrefix + appUsedIn + '. All entries will be deleted.';
```
squallstar: the `Are you sure you want to delete this data source` is repeated twice, please refactor to only append the relevant part.

squallstar: "The data source is currently in use by the following apps: A, B, C."

MaksymShokin: Fixed

MaksymShokin: Fixed

js/interface.js
MaksymShokin: Fixed

PR #66, opened 2019-11-19
When we press ctrl + a key combination we will select only cells with data in them

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/2480

## Description
When we press ctrl + a key combination we will select only cells with data in them

## Screenshots/screencasts
issue-2480

## Backward compatibility

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -17,7 +17,6 @@ var spreadsheet = function(options) {
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
-  
   dataStack = [];
   currentDataStackIndex = 0;
 
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -16,7 +16,6 @@ var spreadsheet = function(options) {
   var dataLoaded = false;
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
-  
   dataStack = [];
   currentDataStackIndex = 0;
 
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -16,7 +16,8 @@ var spreadsheet = function(options) {
   var dataLoaded = false;
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
-
+  var lastKey;
+  
   dataStack = [];
   currentDataStackIndex = 0;
 
@@ -225,6 +226,21 @@ var spreadsheet = function(options) {
     },
     afterSelectionEnd: function(r, c, r2, c2) {
       s = [r, c, r2, c2];
+    },
+    beforeKeyDown: function (event) {
+      if ((lastKey === 17 || lastKey === 91) && event.keyCode === 65 ) {
+        event.stopImmediatePropagation();
+        
+        var colEnd = getColumns().filter(function(column) {
+          return column;
+        }).length - 1;
+        var rowEnd = getData().length;
+
+        hot.deselectCell();
+        hot.selectCells([[0, 0, rowEnd, colEnd]], false, false);
+        return false;
+      }
+      lastKey = event.keyCode;
     }
   };
 
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -249,9 +249,10 @@ var spreadsheet = function(options) {
       if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
         event.stopImmediatePropagation();
         
-        var colEnd = getColumns().filter(function(column) {
+        var cols = getColumns().filter(function(column) {
           return column;
-        }).length - 1;
+        }).length
+        var colEnd = cols ? cols - 1: cols;
         var rowEnd = getData().length;
 
         hot.deselectCell();
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -17,6 +17,7 @@ var spreadsheet = function(options) {
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
+
   dataStack = [];
   currentDataStackIndex = 0;
 
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -17,7 +17,6 @@ var spreadsheet = function(options) {
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
-  var isEditing = false;
 
   dataStack = [];
   currentDataStackIndex = 0;
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -251,7 +251,7 @@ var spreadsheet = function(options) {
         
         var cols = getColumns().filter(function(column) {
           return column;
-        }).length
+        }).length;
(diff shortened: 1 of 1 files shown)


YaroslavOvdii:
@tonytlwu removed the `lastKey` variable. Using `event.ctrlKey` and `event.metaKey` enstead.

js/spreadsheet.js
```diff
@@ -16,7 +16,7 @@ var spreadsheet = function(options) {
   var dataLoaded = false;
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
-
+
```
tonytlwu: No spaces in empty lines.

```suggestion

```

js/spreadsheet.js
```diff
@@ -225,6 +225,22 @@ var spreadsheet = function(options) {
...
+    },
+    beforeKeyDown: function (event) {
+      event = event || window.event;
+
+      if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
+        event.stopImmediatePropagation();
+        
+        var colEnd = getColumns().filter(function(column) {
+          return column;
+        }).length - 1;
```
tonytlwu: What if `.length` returns `0`?

YaroslavOvdii: @tonytlwu 
This scenario is possible only if the user has absolutely empty data source without a single filled row.
And even when a user creates a new one data source it already has default values.
Should we add if condition to check for this scenario?

tonytlwu: My philosophy in error prevention is that when programming, the developer should try to not think about the user scenario but about the possible states of the variable. It isn't useful for developers to say, "well I didn't think users would do XYZ" when a bug occurs.

So yes, we should add conditions to check for this scenario.

tonytlwu: Also, be sure to resolve the conflict before the next review.

YaroslavOvdii:
@tonytlwu added a check for the `colEnd` variable.

js/spreadsheet.js
tonytlwu: Same thing. The spaces in this empty line need to be removed.

js/spreadsheet.js
```diff
@@ -242,6 +242,23 @@ var spreadsheet = function(options) {
...
+    },
+    beforeKeyDown: function (event) {
+      event = event || window.event;
+
+      if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
+        event.stopImmediatePropagation();
+        
+        var cols = getColumns().filter(function(column) {
+          return column;
+        }).length
```
tonytlwu:
```suggestion
        }).length;
```

js/spreadsheet.js
```diff
@@ -242,6 +242,23 @@ var spreadsheet = function(options) {
...
+        var colEnd = cols ? cols - 1: cols;
```
tonytlwu: This would an easier way to understand this.

```suggestion
        var colEnd = Math.max(cols - 1, 0);
```

YaroslavOvdii:
@tonytlwu 
Fixed

js/spreadsheet.js
tonytlwu: This is unresolved

YaroslavOvdii:
@tonytlwu Fixed

js/spreadsheet.js
```diff
@@ -17,7 +17,6 @@ var spreadsheet = function(options) {
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
-
```
tonytlwu: @YaroslavOvdii There should be an empty line after the `var` block.

I'm not sure what's complicated about this change at the moment.

There should be an empty line before and after a `var` block, except at the beginning or end of a clause. This has always been the rule.

An empty line also shouldn't have any spaces and should be completely empty. This has been the issue that I needed addressed, but instead the whole line has been removed. cc @sofiiakvasnevska

js/spreadsheet.js
```diff
@@ -242,6 +242,23 @@ var spreadsheet = function(options) {
     },
     afterSelectionEnd: function(r, c, r2, c2) {
       s = [r, c, r2, c2];
+    },
+    beforeKeyDown: function (event) {
+      event = event || window.event;
+
+      if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
```
tonytlwu: So, this solution looks like when `Ctrl + A` or `Cmd + A` are pressed, all cells are selected.

What happens if the user is currently editing a cell? If that happens, I think users would be expecting for all contents within the cell to be selected like usual spreadsheet behaviour. Does this happen?

YaroslavOvdii: @tonytlwu 

> Does this happen?

No, but I believe I have a solution for this scenario.
We should add an additional variable called `isEditing` with the default value `false`.
In the `handsontable` library, we had a method `beginEditing()` which is calling after a user will begin editing a cell. In this method, we will change `isEditing` variable to `true` and after the user will exit the editing mode. Which will call the `finishEditing()` method, there we will change `isEditing` to the `false` once again.

And alter this method like this:
```
    beforeKeyDown: function (event) {
     if (isEditing){
        return;
      }
     
      event = event || window.event;

      if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
        event.stopImmediatePropagation();
        
        var cols = getColumns().filter(function(column) {
          return column;
        }).length;
        var colEnd = Math.max(cols - 1, 0);
        var rowEnd = getData().length;

        hot.deselectCell();
        hot.selectCells([[0, 0, rowEnd, colEnd]], false, false);
        return false;
      }
    }
```

Will it be a good solution?

YaroslavOvdii: @tonytlwu 
Ignore the previous comment I've found a simpler way.

In the hot object, we have a `getActiveEditor()` method in which we can access the `_opened`
option and it indicates are a user edits a cell or not. It will return `true` or `false`.

So the solution is:
```
    beforeKeyDown: function (event) {
      if (hot.getActiveEditor()._opened){
        return;
      }

      event = event || window.event;

      if ((event.ctrlKey || event.metaKey) && event.keyCode === 65 ) {
        event.stopImmediatePropagation();
        
        var cols = getColumns().filter(function(column) {
          return column;
        }).length;
        var colEnd = Math.max(cols - 1, 0);
        var rowEnd = getData().length;

        hot.deselectCell();
        hot.selectCells([[0, 0, rowEnd, colEnd]], false, false);
        return false;
      }
    }
```

In this case, the user will be able to select a text within the cell while he edits it.

tonytlwu: Yes, that looks like a good recommendation 👍

js/spreadsheet.js
```diff
@@ -17,6 +17,7 @@ var spreadsheet = function(options) {
   var arrayColumns = [];
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
+  var isEditing = false;
```
tonytlwu: Where and how is this variable used?

YaroslavOvdii: @tonytlwu 
That is my mistake. I was trying to implement the previous solution and forgot to remove it.
Sorry about that.

PR #67, opened 2019-11-19
Show to a user message while loading and rendering data

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/3113

## Description
Show to a user message while loading and rendering data

## Screenshots/screencasts
issue-3113

## Backward compatibility

### DIFF
--- a/css/interface.css
+++ b/css/interface.css
@@ -475,3 +475,13 @@ a[data-back] {
     bottom: 60%;
   }
 }
+
+.loading-data {
+  position: fixed;
+  top: 90px;
+  left: 50%;
+  transform: translate(-50%, 0);
+  font-size: 0.9em;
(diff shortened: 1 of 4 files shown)


tonytlwu:
I see there are 2 stages where a loading UI is shown to the user. Is it because the first one is for the API to retrieve data, and the second is to render data?

If so, can we combine them into one single loading UI (the second/new one with the green message)?

YaroslavOvdii:
@tonytlwu 
1) Yes, you are right.

2) I'll add new commit for this PR with needed changes to combine them.

YaroslavOvdii:
@tonytlwu 
Made one loading UI for the component.
Now it looks like that:

new-3113

PR #69, opened 2019-11-26
When a user select any tab except the entries and the quit and return to the data source again he will see the entries tab

YaroslavOvdii:
@sofiiakvasnevska 
## Issue
https://github.com/Fliplet/fliplet-studio/issues/3113

## Description
When a user selects any tab except the entries and the quit and returns to the data source again he will see the entries tab.
Fix the second loader when data source called as an overlay

## Screenshots/screencasts
DSSettings-tab

Fix the second loader when data source called as an overlay
Before
before update

After
after update

## Backward compatibility

## Note

Need to be tested on a large amount of data.

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -330,6 +330,7 @@ function browseDataSource(id) {
   currentDataSourceId = id;
   $contents.addClass('hidden');
   $('.settings-btns').removeClass('active');
+  $initialSpinnerLoading.removeClass('animated');
   // Hide nav tabs and tooltip bar
   var tab = $sourceContents.find('ul.nav.nav-tabs li');
 
@@ -554,6 +555,7 @@ $('#app')
   })
   .on('click', '[data-back]', function(event) {
     event.preventDefault();
+    $('[href="#entries"]').click();
     if (!dataSourceEntriesHasChanged || confirm('Are you sure? Changes that you made may not be saved.')) {
       try{
         table.destroy();
--- a/js/interface.js
+++ b/js/interface.js
@@ -339,10 +339,10 @@ function browseDataSource(id) {
     }
   });
 
+  $('[href="#entries"]').click();
   $sourceContents.find('#toolbar').hide();
   $('.loading-data').show();
   $sourceContents.removeClass('hidden');
-  $('.entries-message').html('<br>Loading data...');
 
   // Input file temporarily disabled
   // $contents.append('<form>Import data: <input type="file" /></form><hr /><div id="entries"></div>');
@@ -353,7 +353,6 @@ function browseDataSource(id) {
       fetchCurrentDataSourceDetails()
     ])
     .then(function() {
-      $('[href="#entries"]').click();
       windowResized();
 
       if (copyData.context === 'overlay') {
@@ -875,7 +874,10 @@ $('#app')
         } catch(e) {}
       }
     } else {
-      hot.render();
+      if (hot.container !== null) {
+        hot.render();
+      }
+
       $('.save-btn').removeClass('hidden');
       $('.back-name-holder').removeClass('hide-date');
       $('.controls-wrapper').removeClass('data-settings data-roles');
--- a/js/interface.js
+++ b/js/interface.js
@@ -1,4 +1,4 @@
-var $initialSpinnerLoading = $('.spinner-holder.inital-state');
+var $initialSpinnerLoading = $('.loading-data');
 var $contents = $('#contents');
 var $sourceContents = $('#source-contents');
 var $dataSources = $('#data-sources > tbody');
@@ -34,7 +34,7 @@ var definitionEditor = CodeMirror.fromTextArea($('#definition')[0], {
 
 // Fetch all data sources
 function getDataSources() {
-  $initialSpinnerLoading.addClass('animated');
+  $initialSpinnerLoading.show();
   $contents.addClass('hidden');
   $sourceContents.addClass('hidden');
   $('[data-save]').addClass('hidden');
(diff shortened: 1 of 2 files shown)


tonytlwu:
Can we remove the first type of spinning loader so that we're only using the new loader? It doesn't look great having 2 different types of loaders.

YaroslavOvdii:
@tonytlwu 
The first loader is using also when we await until the data sources list is loaded.
And using there new loading UI will not look great.

Here is a place I was talking about.
first-loader

If we remove the first loader user will not know what is happening while the data source list is loading because we won't show anything to him.

tonytlwu:
By "remove" I meant stop using that style of loader. We should just use one single loader UI design, and never show two loaders on screen at the same time.

Loading from outside an app

1. Show green loader when data source list is loading
1. Show green loader when data source entries are loading

Loading from inside an app in an overlay

1. Show green loader when both the data source list and data source entries are loading
1. Don't show one loader one top of the another

All of this is to show the user only one UI that communicates, "something is loading".

YaroslavOvdii:
@tonytlwu I get it now. I'll do a new commit for this PR with these changes.

YaroslavOvdii:
@tonytlwu 
It will look like that.
new-loader

Will it be okay?

tonytlwu:
Yup. That looks better.

YaroslavOvdii:
@tonytlwu added these changes to the PR

js/interface.js
```diff
@@ -554,6 +554,7 @@ $('#app')
   })
   .on('click', '[data-back]', function(event) {
     event.preventDefault();
+    $('[href="#entries"]').click();
```
tonytlwu: Rather than "switch to the Entries tab before leaving a data source view and going to the list view", we should be implementing "when going to a data source view, make sure it's the Entries tab".

The reason is that the action of "opening Entries tab" should be associated with the "opening a data source" action. If anything was changed about how we go back to the list view, there's a risk of creating unexpected behaviours since it's not an expectation to associate "leaving the data source view" with any kind of tab switching.

YaroslavOvdii: @tonytlwu 
The idea to switch tabs comes to me from the example of how we [update](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L774) setting.
As you may see we are already doing tab switch on [saving settings](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L791).

The issue will occur only when we try to view a large data source, because of the UI freeze by the time we get to this [point](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L356).
In another case, the user won't see the settings tab because we will go to it when we load the data source before it will be rendered.

And if we try to set entries tab before data source entries are loaded we receive an error in [this](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L877) part of the code.

image

Could you please guide me in this situation? 
Is it a good idea to check if this `handsontable` instance has been destroyed and do not call `hot.render()` method?

YaroslavOvdii: @tonytlwu I've found how we can work around this issue and not clicking a entries tab when we are returning to the list of the data sources.

This [line](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L877) stopped us from clicking to the entries tab right away after we enter the data source.
To avoid error in this place we could simply check if there is a `container` option in the `hot` object.
If it isn't equal `null` these will mean that `handsontable` lib was initialized already and we can call a `hot.render()` method. If it is null then we just skip the `hot.render()` method call.
In this case, we just click the entires tab right away when we open the data source, somewhere at this [point](https://github.com/Fliplet/fliplet-widget-data-source/blob/b710b29127e1a0280b38d473c724fc63ba58a014/js/interface.js#L341) and the user will always see the entries tab when he opens the data source.

This is an [example](https://github.com/YaroslavOvdii/fliplet-widget-data-source/commit/d30983dbd4b32be3e5f9fa65b7ad0cb37076d64d) of changes I suggesting.

tonytlwu: @YaroslavOvdii I didn't realise that there these many complications with opening the Entries tab when the data source is opened.

I understand that you're checking if `hot.container` is `null` before you run `hot.render()`. However what happens if we don't run `hot.render()`? Does the table not render? How are we making sure the table rendering is automatically started when users open a data source. We don't users to have to take an extra action after they open the data source.

YaroslavOvdii: > However what happens if we don't run hot.render()? Does the table not render?

Yes, the table won't render, and we don't want it to render at that point, because the hot object wasn't initialized yet.
However, it will render after the data will be loaded from the backend with this [method](https://github.com/YaroslavOvdii/fliplet-widget-data-source/blob/d30983dbd4b32be3e5f9fa65b7ad0cb37076d64d/js/interface.js#L195).

> How are we making sure the table rendering is automatically started when users open a data source. We don't users to have to take an extra action after they open the data source.

Users will not have to take any extra action after they open the data source because we will take care of the switching tab for them at [this](https://github.com/YaroslavOvdii/fliplet-widget-data-source/blob/d30983dbd4b32be3e5f9fa65b7ad0cb37076d64d/js/interface.js#L342) place.

All we have done is simply moved the switch to the entries tab before data is loaded from the backend. This will allow as to switch to the entries tab before UI will freeze because of the rendering the table. In this way, a user will always see the Entries tab when he opens the data source.

> I understand that you're checking if hot.container is null before you run hot.render().

And this part is needed because of this issue.

>And if we try to set entries tab before data source entries are loaded we receive an error in this part of the code.

tonytlwu: Thanks for the detailed explanation. Sounds like a good way to go 👍

PR #71, opened 2020-02-06
Error catch on commit save to display a modal

squallstar:
image


js/interface.js
```diff
@@ -580,7 +580,17 @@ $('#app')
       }
 
       $('.data-save-updated').html('All changes saved!');
-    })
+    }).catch(function (err) {
+      Fliplet.Modal.alert({
+        message: Fliplet.parseError(err)
```
tonytlwu:
```suggestion
        title: 'Error saving data source',
        message: Fliplet.parseError(err)
```

PR #75, opened 2020-03-30
When user press ctrl+a combination selection will be correct

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6108

## Description
When user press ctrl+a combination selection will be correct

## Screenshots/screencasts
https://share.getcloudapp.com/YEuAEeEq

## Backward compatibility

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -68,54 +68,71 @@ var spreadsheet = function(options) {
   }
 
   function getNearestCells(cellPosition) {
-    var leftShift = cellPosition[1] - 1;
-    var topShift = cellPosition[0] - 1;
-    var left = [cellPosition[0], leftShift < 0 ? cellPosition[1] : leftShift];
-    var right = [cellPosition[0], cellPosition[1] + 1];
-    var top = [topShift < 0 ? cellPosition[0] : topShift, cellPosition[1]];
-    var bottom = [cellPosition[0] + 1, cellPosition[1]];
+    var leftShift = cellPosition.y - 1;
+    var topShift = cellPosition.x - 1;
+    var left = {
+      x: cellPosition.x,
+      y: leftShift < 0 ? cellPosition.y : leftShift
+    };
+    var right = {
+      x: cellPosition.x,
+      y: cellPosition.y + 1
+    };
+    var top = {
+      x: topShift < 0 ? cellPosition.x : topShift,
+      y: cellPosition.y
+    };
+    var bottom = {
+      x: cellPosition.x + 1,
+      y: cellPosition.y
+    };
 
-    return [left, right, top, bottom];
+    return {
+      left: left,
+      right: right,
+      top: top,
+      bottom: bottom 
+    };
   }
 
   /**
    * The function that surfs through the table data for detecting closest cells with data from the selected cell
    * 
    * @param {array} tableData - an array of table data received from the hot.getData()
-   * @param {array} cellPosition - an array of the cell coordinates
+   * @param {object} cellPosition - an object of the cell coordinates with x and y keys
    * @param {array} processedCells - an array of the cells we have already checked
    * @returns {array} returns array of the scaned cells example array[rowIndex][cellIndex] = hasData true/false
    */
   function getDataCoordinates(tableData, cellPosition, processedCells) {
     if (!processedCells) {
       processedCells = [];
-      processedCells[cellPosition[0]] = [];
-      processedCells[cellPosition[0]][cellPosition[1]] = true;
+      processedCells[cellPosition.x] = [];
+      processedCells[cellPosition.x][cellPosition.y] = true;
     }
 
     var cellsGroup = getNearestCells(cellPosition);
 
-    cellsGroup.push(cellPosition);
+    cellsGroup['currentCell'] = cellPosition;
 
-    cellsGroup.forEach(function(cell) {
-      var hasData = tableData[cell[0]][cell[1]] !== null;
-      var processedRow = processedCells[cell[0]];
+    for (var cell in cellsGroup) {
+      var hasData = tableData[cellsGroup[cell].x][cellsGroup[cell].y] !== null;
+      var processedRow = processedCells[cellsGroup[cell].x];
 
       if (!processedRow) {
-        processedCells[cell[0]] = [];
+        processedCells[cellsGroup[cell].x] = [];
       }
 
-      var processedCell = processedCells[cell[0]][cell[1]];
+      var processedCell = processedCells[cellsGroup[cell].x][cellsGroup[cell].y];
 
       if (!processedCell) {
-        processedCells[cell[0]][cell[1]] = false;
+        processedCells[cellsGroup[cell].x][cellsGroup[cell].y] = false;
       }
 
       if (hasData && (!processedRow || !processedCell)) {
-        processedCells[cell[0]][cell[1]] = true;
-        getDataCoordinates(tableData, cell, processedCells);
-      }
-    });
+        processedCells[cellsGroup[cell].x][cellsGroup[cell].y] = true;
+        getDataCoordinates(tableData, cellsGroup[cell], processedCells);
+      }      
+    }
 
     return processedCells;
   }
@@ -153,7 +170,11 @@ var spreadsheet = function(options) {
    * @returns {array} range of the data that we need to select
    */
   function getSelectionRange(selectionPosition) {
-    var dataCoordinates = getDataCoordinates(hot.getData(), selectionPosition[0]);
+    var selectedPosition = {
+      x: selectionPosition[0][0],
+      y: selectionPosition[0][1]
+    };
+    var dataCoordinates = getDataCoordinates(hot.getData(), selectedPosition);
     var selectedCells = 0;
 
     dataCoordinates.forEach(function(row) {
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -94,6 +94,7 @@ var spreadsheet = function(options) {
     }
 
     var cellsGroup = getNearestCells(cellPosition);
+
     cellsGroup.push(cellPosition);
 
     cellsGroup.forEach(function(cell) {
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -68,247 +68,108 @@ var spreadsheet = function(options) {
   }
 
   /**
-   * We user this method to determine where is closest data is lokated from the selected cell
+   * The function that surfs through the table data for detecting closest cells with data from the selected cell
    * 
-   * @param {Array} selectedCell - array of the coordinat of the selected cell reacived throw the hot.getSelected() method
-   * 
-   * @returns {Object} - object of the directions where data is placed occording to the selected cell
+   * @param {array} tableData - an array of table data received from the hot.getData()
+   * @param {array} cellPosition - an array of the cell coordinates
+   * @param {array} processedCells - an array of the cells we have already checked
+   * @returns {array} returns array of the scaned cells example array[rowIndex][cellIndex] = hasData true/false
    */
-  function closestData(selectedCell) {
-    selectedCell = selectedCell[0];
-    if (!Array.isArray(selectedCell)) {
-      console.error('We must pass an array of the cell coordinats to the closestData function. First element is cell' +
-        'row and second element is cell col. In this case script will act as if there was a value in the cell. ' +
-        'Value that was passed - ',
-        selectedCell);
-      return false;
+  function getDataCoordinates(tableData, cellPosition, processedCells) {
+    if (!processedCells) {
+      processedCells = [];
+      processedCells[cellPosition[0]] = [];
+      processedCells[cellPosition[0]][cellPosition[1]] = true;
     }
 
-    var col = selectedCell[1];
-    var row = selectedCell[0];
-    var selectedCellData = hot.getDataAtCell(row, col)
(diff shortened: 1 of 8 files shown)


js/spreadsheet.js
```diff
@@ -67,248 +67,109 @@ var spreadsheet = function(options) {
...
+  function getNearestCells(cellPosition) {
+    var leftShift = cellPosition[1] - 1;
+    var topShift = cellPosition[0] - 1;
+    var left = [cellPosition[0], leftShift < 0 ? cellPosition[1] : leftShift];
+    var right = [cellPosition[0], cellPosition[1] + 1];
+    var top = [topShift < 0 ? cellPosition[0] : topShift, cellPosition[1]];
+    var bottom = [cellPosition[0] + 1, cellPosition[1]];
 
-    return dataAt;
+    return [left, right, top, bottom];
```
tonytlwu, line 78: Could we not be more specific like this in all this new code? The current format is more complex to read, understand and maintain.

```suggestion
    return {
      top: top,
      right: right,
      bottom: bottom,
      left: left
    };
```

js/spreadsheet.js
```diff
@@ -67,248 +67,109 @@ var spreadsheet = function(options) {
...
+      processedCells[cellPosition[0]] = [];
+      processedCells[cellPosition[0]][cellPosition[1]] = true;
+    }
 
-      lastRow = _.max([lastRow - 1, 0]);
-    } else if (moveTo.right) {
-      // When data located on the right of the selected cell
-      firstCol = startAt[1];
+    var cellsGroup = getNearestCells(cellPosition);
+    cellsGroup.push(cellPosition);
```
tonytlwu, line 97: Check for all linting errors.

```suggestion

    cellsGroup.push(cellPosition);
```

js/spreadsheet.js
```diff
@@ -67,248 +67,109 @@ var spreadsheet = function(options) {
...
+  /**
+   * The method that returns us a range that we should select in the table
+   * 
+   * @param {array} processedCells array processed by getDataCoordinates function
+   */
+  function getSelectionCoordinates(processedCells) {
+    var rowIndexes = [];
+    var colIndexes = [];
+
+    processedCells.forEach(function(row, rowIndex) {
```
tonytlwu, line 131:
```suggestion

    processedCells.forEach(function(row, rowIndex) {
```

js/spreadsheet.js
YaroslavOvdii: @tonytlwu We could do it in this function, yes, but not for this [function](https://github.com/Fliplet/fliplet-widget-data-source/pull/75/files#diff-58f9b9f8183cba2a53b4728f2df11588R127) because such format needed for the `handsontable` library.

js/spreadsheet.js
YaroslavOvdii: @tonytlwu do we need one more new line here?

js/spreadsheet.js
YaroslavOvdii: @tonytlwu also about this return we decided to use this format:
1. It is easy to expand functionality for this feature when we decide to look for data in diagonal.
2. This return used in a sequence of the functions we have written and we are expecting an array to be a result of this function. So if we change it to the object we will have to transform it back to the array after that

js/spreadsheet.js
tonytlwu: No we don't :) looks like there's one already.

js/spreadsheet.js
tonytlwu: 1. What do you mean by deciding to look for data in diagonal? Can you give me code examples of what it would roughly look in one method compared to another?
2. I think if we use `.top` and other properties, it's essentially only in this line https://github.com/Fliplet/fliplet-widget-data-source/pull/75/files#diff-58f9b9f8183cba2a53b4728f2df11588R369 that it needs to be converted to an array format (because handsontable needs it). I don't like the fact that all these small functions are written in a format handsontable dictates when it can just be a helper function that transforms these in the moment right before handsontable needs them. (We don't necessarily have to change the format now, as I'd like to see what you mean by point 1 first to decide.)

YaroslavOvdii: > What do you mean by deciding to look for data in diagonal? Can you give me code examples of what it would roughly look in one method compared to another?

By diagonal I mean that now we look for data on top, left, bottom and right to decide should we what range we should select. But if we will want to look for data at top-left, top-right bottom-left and bottom-right in the current function we will simply add these cells to the array but with the suggested function, we will need to add an extra object key and search for it in next functions.
For example changes with our function:
```
  function getNearestCells(cellPosition) {
    var leftShift = cellPosition[1] - 1;
    var topShift = cellPosition[0] - 1;
    var left = [cellPosition[0], leftShift < 0 ? cellPosition[1] : leftShift];
    var right = [cellPosition[0], cellPosition[1] + 1];
    var top = [topShift < 0 ? cellPosition[0] : topShift, cellPosition[1]];
    var bottom = [cellPosition[0] + 1, cellPosition[1]];
    var topLeft = [topShift < 0 ? cellPosition[0] : topShift, leftShift < 0 ? cellPosition[1] : leftShift];
    var topRight = [cellPosition[0] + 1, cellPosition[1]];

    return [left, right, top, bottom, topLeft, topRight ];
  }
```

Changes with an object for top value only.

```
  function getNearestCells(cellPosition) {
    var leftShift = cellPosition[1] - 1;
    var topShift = cellPosition[0] - 1;
    var left = [cellPosition[0], leftShift < 0 ? cellPosition[1] : leftShift];
    var right = [cellPosition[0], cellPosition[1] + 1];
    var top = [topShift < 0 ? cellPosition[0] : topShift, cellPosition[1]];
    var bottom = [cellPosition[0] + 1, cellPosition[1]];
    var topLeft = [topShift < 0 ? cellPosition[0] : topShift, leftShift < 0 ? cellPosition[1] : leftShift];
    var topRight = [cellPosition[0] + 1, cellPosition[1]];

    return {
      left: left,
      right: right,
      top: top,
      bottom: bottom,
      topLeft: topLeft,
      topRight: topRight
    };
  }

  function getDataCoordinates(tableData, cellPosition, processedCells) {
    if (!processedCells) {
      processedCells = [];
      processedCells[cellPosition[0]] = [];
      processedCells[cellPosition[0]][cellPosition[1]] = true;
    }

    var cellsGroup = getNearestCells(cellPosition);

    cellsGroup['currentCell'] = cellPosition;

    var processedRow;
    var processedCell;

    if (tableData[cellsGroup.top] !== null) {
      processedRow = processedCells[cellsGroup.top[0]];

      if (!processedRow) {
        processedCells[cellsGroup.top[0]] = [];
      }

      processedCell = processedCells[cellsGroup.top[0]][cellsGroup.top[1]];

      if (!processedCell) {
        processedCells[cellsGroup[cell][0]][cellsGroup[cell][1]] = false;
      }
      
      getDataCoordinates(tableData, cellsGroup.top, )
    }

    // And for every key in the object, we will need such if
}
```

Example with `for ... in` : 
```
  function getDataCoordinates(tableData, cellPosition, processedCells) {
    if (!processedCells) {
      processedCells = [];
      processedCells[cellPosition[0]] = [];
      processedCells[cellPosition[0]][cellPosition[1]] = true;
    }

    var cellsGroup = getNearestCells(cellPosition);

    cellsGroup['currentCell'] = cellPosition;

    for (var cell in cellsGroup) {
      var hasData = tableData[cellsGroup[cell][0]][cellsGroup[cell][1]] !== null;
      var processedRow = processedCells[cellsGroup[cell][0]];

      if (!processedRow) {
        processedCells[cellsGroup[cell][0]] = [];
      }

      var processedCell = processedCells[cellsGroup[cell][0]][cellsGroup[cell][1]];

      if (!processedCell) {
        processedCells[cellsGroup[cell][0]][cellsGroup[cell][1]] = false;
      }

      if (hasData && (!processedRow || !processedCell)) {
        processedCells[cellsGroup[cell][0]][cellsGroup[cell][1]] = true;
        getDataCoordinates(tableData, cell, processedCells);
      }      
    }
```

tonytlwu: @YaroslavOvdii I think I'm understanding this a little better now. Much of the confusion is created by the `[0]`, `[1]` notations and ambiguous naming, which made it difficult to understand.

Am I right in assuming that `getNearestCells()` takes the position of one cell and returns a range of cells, e.g. from A1 to E5? If so, then:

1. I still recommend using objects and properties instead of an array of numbers, which is unclear.
1. The range would be defined by either:
    1. `{ topLeft, topRight, bottomRight, bottomLeft }` or
    1. `{ top, left, columns, rows }` **Preferred**

YaroslavOvdii: @tonytlwu 
> Am I right in assuming that getNearestCells() takes the position of one cell and returns a range of cells, e.g. from A1 to E5? If so, then:

No, it is not a range. It's simply coordinates of the four cells around one cell. And we won't be able to refuse from the `[0]` and `[1]` because the cell position is an array `[rowIndex, colIndex]` and we needed this coordinates to find the value of the cell in the `tableData` array.
I'll try to remade as

tonytlwu: @YaroslavOvdii I see. Then I think `{ top, left, bottom, right }` is the right way to go. If "top left" was needed, I'd just take the x coordinate of one and y coordinate of the other when it is needed, without needing to provide additional properties.

YaroslavOvdii: @tonytlwu 
If you don't mind I would describe a full flow here.

The main idea was that we will serf threw the cells of the data source to determine a range of data we need to select.

Step 1.
Get selected cell coordinates.
image

Step 2.
Get the nearest cells top, bottom, left, right.
image

Step 3.
Find if there is a data in the cells nearby to the selected cell.
If we have data in any cell from Step 2, we set it coordinates for Step 1 and repeat these steps until we won't get all cells with data scanned, but with a condition that it doesn't have to be a space between cells with data.
So in our case scan will go left, right, but won't go top and bottom because there is no data in those cells.

Step 4.
When we get a scanned cells array which looks like this `[rowIndex][cellIndex] = hasData true/false`  we simply get `max` and `min` values of the `rowsIndex` and `cellIndex`  which has `true` that is our range that we need to select for the user.

YaroslavOvdii: @tonytlwu added changes as suggested.

js/spreadsheet.js
```diff
@@ -67,248 +67,131 @@ var spreadsheet = function(options) {
...
+      processedCells[cellPosition.x] = [];
+      processedCells[cellPosition.x][cellPosition.y] = true;
+    }
 
-      for (var i = startAt[0]; i > 0; i--) {
-        if (allData[i][lastCol] === null) {
-          firstRow = i ? i - 1 : i;
-        }
-      }
+    var cellsGroup = getNearestCells(cellPosition);
```
tonytlwu, line 113:
```suggestion
    var nearestCells = getNearestCells(cellPosition);
```

js/spreadsheet.js
```diff
@@ -67,248 +67,131 @@ var spreadsheet = function(options) {
...
 
-      firstRow = firstRow || 0;
+    cellsGroup['currentCell'] = cellPosition;
```
tonytlwu, line 115: Why would there be a `currentCell` property? I don't see it in the `getNearestCells()` function.

js/spreadsheet.js
```diff
@@ -67,248 +67,131 @@ var spreadsheet = function(options) {
...
-          lastCol = i - 1;
-          break;
-        }
-      }
-    } else if (moveTo.bottom) {
-      // When data located on the bottom of the selected cell
-      firstRow = startAt[0];
+      if (hasData && (!processedRow || !processedCell)) {
+        processedCells[cellsGroup[cell].x][cellsGroup[cell].y] = true;
+        getDataCoordinates(tableData, cellsGroup[cell], processedCells);
```
tonytlwu, line 133: Am I right in assuming that the process of selecting all cells is a recursive process, as it's indicated in this line?

If so, how efficient is it if we want to select all 100K rows and 100 columns?

js/spreadsheet.js
```diff
@@ -67,248 +67,131 @@ var spreadsheet = function(options) {
...
-        if (allData[lastRow][i] === null) {
-          firstCol = i + 1;
-          break;
+  /**
+   * The method that returns us a range that we should select in the table
+   * 
+   * @param {array} processedCells array processed by getDataCoordinates function
+   */
+  function getSelectionCoordinates(processedCells) {
+    var rowIndexes = [];
```
tonytlwu, line 146: The plural for **index** is **indices**.

```suggestion
    var rowIndices = [];
```

js/spreadsheet.js
YaroslavOvdii: @tonytlwu 
`currentCell` needed for when our selected cell is without a data on it row or col.

tonytlwu: @YaroslavOvdii Let's just add it into `getNearestCells()`. The `currentCell` property is 100% always added at this point, and it's good for the nearest cells to have a reference of which cell was the starting point.

js/spreadsheet.js
YaroslavOvdii: @tonytlwu You were right about large amount of data. This couldn't handle it. I've created a new [PR](https://github.com/Fliplet/fliplet-widget-data-source/pull/78) with a new solution and tested it on a large data source 60k rows.

And I'll close this PR because it is not needed anymore.

PR #76, opened 2020-03-30
Support for Version history

squallstar:
Todo (all done)

- ~UI WIP~
- ~There's a few bugs to iron out~

---

# UI

Full flow https://share.getcloudapp.com/04uKjk4R

## List of versions

image

---

## Previewing a version

image

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -237,6 +237,8 @@ function fetchCurrentDataSourceEntries(entries) {
         }];
         columns = ['Column 1', 'Column 2'];
       } else {
+        $('#show-versions').show();
+
         // Let's make sure we get all the columns checking all rows
         // and add any missing column to the datasource columns
         rows.forEach(function addMissingColumns(row) {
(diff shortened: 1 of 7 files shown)


js/interface.js
```diff
@@ -909,9 +1002,65 @@ $('#app')
...
+  .on('click', '[data-version-copy]', function (e) {
+    e.preventDefault();
+    var id = $(this).data('version-copy');
+
+    return createDataSource({
+      version: {
+        dataSourceId: currentDataSourceId,
+        id: id
+      }
+    }).then(function () {
```
tonytlwu: Can we pass to this function `Copy of $dataSourceName` as the default name? Users can then change it in the prompt.

squallstar: done

js/interface.js
```diff
@@ -478,11 +478,13 @@ function browseDataSource(id) {
...
 }
 
-function createDataSource(createOptions) {
+function createDataSource(createOptions, options) {
   createOptions = createOptions || {};
+  options = options || {};
 
   return Fliplet.Modal.prompt({
     title: 'Enter the name of your new Data Source',
+    value: _.get(options, 'value', '')
```
tonytlwu: Semantically, the option provided is for the data source name, not for data source value.

```suggestion
    value: _.get(options, 'name', '')
```

templates/versions.interface.hbs
```diff
@@ -22,11 +22,17 @@
...
-              <li><a href="#" data-version-preview="{{id}}">Preview data</a></li>
-              <li><a href="#" data-version-restore="{{id}}">Restore</a></li>
-              <li><a href="#" data-version-copy="{{id}}">Copy to new data source</a></li>
-            </ul>
+            <div class="btn-group" role="group">
+              <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                Actions
+                <span class="caret"></span>
+              </button>
+              <ul class="dropdown-menu">
```
tonytlwu: With the example on https://getbootstrap.com/docs/3.4/javascript/#dropdowns, if you add the following styles to the `UL`, it'll align the dropdown to the right edge. You won't need to set the actions column with a fixed width. This assumes `.btn-group` has `position: relative`.

```css
    left: auto;
    right: 0;
```

js/interface.js
squallstar: done

templates/versions.interface.hbs
squallstar: thanks, works 
image

PR #78, opened 2020-04-09
When user press ctrl+a combination selection will be correct

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/6108

## Description
When user press ctrl+a combination selection will be correct

## Screenshots/screencasts
https://share.getcloudapp.com/OAub0Y6m
https://share.getcloudapp.com/Z4uLbO18 - large data selection 60k rows ~100 cols

## Backward compatibility

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -68,243 +68,195 @@ var spreadsheet = function(options) {
   }
 
   /**
-   * We user this method to determine where is closest data is lokated from the selected cell
+   * Function to check is there any data in the nearest cells
    * 
-   * @param {Array} selectedCell - array of the coordinat of the selected cell reacived throw the hot.getSelected() method
+   * @param {object} selectedCell - object with keys row and col which contains selected cell coordinates
+   * @param {array} tableData - an array of the table data from Handsontable library
    * 
-   * @returns {Object} - object of the directions where data is placed occording to the selected cell
+   * @returns {object} - possible keys: right, left, top, bottom - shows which closest cells have data
    */
-  function closestData(selectedCell) {
-    selectedCell = selectedCell[0];
-    if (!Array.isArray(selectedCell)) {
-      console.error('We must pass an array of the cell coordinats to the closestData function. First element is cell' +
-        'row and second element is cell col. In this case script will act as if there was a value in the cell. ' +
-        'Value that was passed - ',
-        selectedCell);
-      return false;
-    }
-
-    var col = selectedCell[1];
-    var row = selectedCell[0];
-    var selectedCellData = hot.getDataAtCell(row, col)
-    // At this block we getting an index of the nearest cells from the selected cell
-    // If we selected first row it ID is 0 already and if we - 1 from it we will reacive an error in the hot.getDataAtCell() method
-    var top = row ? row - 1 : row;
-    var bottom = row + 1;
-    // Same as in top variable
-    var left = col ? col - 1 : col;
-    var right = col + 1;
-    // At this block we reacive the nearest cells value from the selected cell
-    var leftValue = hot.getDataAtCell(row, left);
-    var rightValue = hot.getDataAtCell(row, right);
-    var topValue = hot.getDataAtCell(top, col);
-    var bottomValue = hot.getDataAtCell(bottom, col);
-    var dataAt = {
-      left: false,
-      right: false,
-      top: false,
-      bottom: false,
-      all: false,
-      hasData: false
+  function nearestCellsWithData(selectedCell, tableData) {
+    var cellsWithData = {
+      right: tableData[selectedCell.row][selectedCell.col + 1] !== null,
+      left: tableData[selectedCell.row][selectedCell.col - 1] !== null,
+      top: tableData[selectedCell.row - 1][selectedCell.col] !== null,
+      bottom: tableData[selectedCell.row + 1][selectedCell.col] !== null,
     };
+    var countCellsWithData = 0;
 
-    // If there is a data in the selected cell we should select data releated to this cell
-    if (selectedCellData !== null) {
-      dataAt.hasData = true;
-      return dataAt;
+    for (var key in cellsWithData) {
+      if (cellsWithData[key]) {
+        countCellsWithData++;
+      }
+    }
+    
+    // If there is no true value in cellsWithData object we should select all table
+    if (countCellsWithData === 0) {
+      return {selectAll: true};
     }
 
-    // If no value near the selected cell we should select all table, also fires when we just load data source
-    // and clicked ctrl+a combination
-    if (leftValue === null && rightValue === null && topValue === null && bottomValue === null) {
-      dataAt.all = true;
-      return dataAt;
+    if (tableData[selectedCell.row][selectedCell.col] !== null) {
+      return {cellWithData: true};
     }
-
-    // Showing where is data position from the selected cell
-    dataAt.left = leftValue !== null;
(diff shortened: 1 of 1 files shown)


js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
-    var topValue = hot.getDataAtCell(top, col);
-    var bottomValue = hot.getDataAtCell(bottom, col);
-    var dataAt = {
-      left: false,
-      right: false,
-      top: false,
-      bottom: false,
-      all: false,
-      hasData: false
+  function nearestCellsWithData(selectedCell, tableData) {
```
tonytlwu, line 78:
```suggestion
  function getAdjacentCellsWithData(selectedCell, tableData) {
```

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
+    var cellsWithData = {
+      right: tableData[selectedCell.row][selectedCell.col + 1] !== null,
+      left: tableData[selectedCell.row][selectedCell.col - 1] !== null,
+      top: tableData[selectedCell.row - 1][selectedCell.col] !== null,
+      bottom: tableData[selectedCell.row + 1][selectedCell.col] !== null,
```
tonytlwu, line 83: Add a `current: tableData[selectedCell.row][selectedCell.col] !== null`, then just return the `cellsWithData` object. In fact, it doesn't really a `var` declared.

The reason for this is that the function will do exactly what the name suggests, and does not add logic on top to dictate what other functions should do with the information. The caller of this function can then decide under what scenario it should _select all_ etc.

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
+    } else {
+      for (currentIndex; currentIndex >= 0; currentIndex--) {
+        if (tableData[currentIndex] === null) {
           break;
         }
       }
+    }
 
-      firstRow = firstRow || 0;
+    return currentIndex < 0 ? 0: currentIndex;
```
tonytlwu, line 130:
```suggestion
    return Math.max(0, currentIndex);
```

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
 
-      firstRow = firstRow || 0;
+  /**
+   * Function to get coordinates which we need to select
+   * 
+   * @param {Array} cellPosition - an array of the selected coordinates
+   * 
+   * @returns {Array/Boolean} - coordinates that need to be selected. Example of the returned data: [[startRow, startCol, endRow, endCol]]
+   */
+  function coordinatesToSelect(cellPosition) {
```
tonytlwu, line 169:
```suggestion
  function getSelectAllCoordinates(cellPosition) {
```

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
+    var nearestDataAt = nearestCellsWithData(selectedCellPosition, tableData);
 
-      for (var i = lastRow - 1; i > 0; i--) {
-        if (allData[i][startAt[1]] === null) {
-          firstRow = i;
-          break;
-        }
-      }
+    // Return false in the case when we need to select all table data
+    if (nearestDataAt.selectAll) {
```
tonytlwu, line 182: Move the `for (var key in ...)` logic from inside the `getAdjacentCellsWithData()` function to here.

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
-   * Method to get a coordinats which we need to select
-   * 
-   * @param {Array} startAt - array of the selected coordinats 
-   * @param {Object} moveTo - object that returned from closestData() function
+   * The function that helps us to find start index for further selection
    * 
-   * @returns {Array} - coordinats that needs to be selected. Example of the returned data: [[startRow, startCol, endRow, endCol]]
+   * @param {array} tableData - an array of data from the Handsontable library
+   * @param {int} startSearchFrom - index of the row/col from what we should start a search
+   * @param {int} colIndex - a not necessary option that tells us what we need to look, a row index or column index
```
tonytlwu, line 110: Am I right in suggesting this?

```suggestion
   * @param {int} colIndex - (optional) when provided, the search is performed along the rows for the column specified
```

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
-        if (allData[firstRow][i] === null) {
-          firstCol = i ? i + 1 : i;
-          break;
-        }
-      }
+      firstCol = getStartIndex(tableData[selectedCellPosition.row], col);
 
-      firstCol = firstCol || 0;
+      // If column index is not 0 it means that data what we need to select is in the next column
+      firstCol = firstCol === 0 ? firstCol : firstCol + 1;
```
tonytlwu, line 192: If I'm reading this right, if `firstCol` is `1`, you'd end up adding `1` to it and make it `2`? What's the reason behind doing this?

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
+      firstRow = getStartIndex(tableData, selectedCellPosition.row, firstCol);
 
-      for (var i = firstRow + 1; i < allData.length; i++) {
-        if (allData[i][startAt[1]] === null) {
-          lastRow = i - 1;
-          break;
-        }
-      }
+      // If row index is not 0 it means that data what we need to select is in the next row
+      firstRow = firstRow === 0 ? firstRow : firstRow + 1;
```
tonytlwu, line 197: Same as above. Why are we turning `1` to `2` while `0` remains the same as `0`?

js/spreadsheet.js
```diff
@@ -68,244 +68,196 @@ var spreadsheet = function(options) {
...
 
-      for (var i = startAt[1]; i > 0; i--) {
-        if (allData[lastRow][i] === null) {
-          firstCol = i + 1;
-          break;
-        }
-      }
+      lastCol = getEndIndex(tableData[firstRow], firstCol);
+      lastCol = lastCol < selectedCellPosition.col ? selectedCellPosition.col : lastCol;
```
tonytlwu, line 200:
```suggestion
      lastCol = Math.min(lastCol, selectedCellPosition.col);
```

js/spreadsheet.js
YaroslavOvdii: @tonytlwu it would be easier to explain on an example:
How selection will behave without this line.
https://share.getcloudapp.com/yAu2Pode

And with this line.
https://share.getcloudapp.com/WnuG1woX

This behavior occurs because of how we searching for the coordinates. 
Our functions `getStartIndex` and `getEndIndex` are looking for the `null` in value of the cell to determine that there is no data in that cell. That is why they are returning an index of the cell with null, but we are interested to select only cells with data, that is why we adding to the row or cell 1 here.

js/spreadsheet.js
YaroslavOvdii: @tonytlwu, in this case, we should use `Math.max` as well. Because `Math.min` will cause this issue https://share.getcloudapp.com/wbu0AeAb

tonytlwu:
@YaroslavOvdii @sofiiakvasnevska What's happening with this? Looks like it's been put on pause. Can we continue with this?

PR #79, opened 2020-04-14
When user sort data sources by the name they will be sorted correctly

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5318

## Description
When user will sort data sources by the name they will be sorted correctly

## Screenshots/screencasts
https://share.getcloudapp.com/v1uD6yrZ

## Backward compatibility

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -575,18 +575,18 @@ function sortDataSources(key, order) {
     toBeOrderedDataSources = allDataSources;
   }
 
-  // Order by updatedAt
   var orderedDataSources = _.orderBy(toBeOrderedDataSources, function(ds) {
-    var date = new Date(ds[key]);
-    if (Object.prototype.toString.call(date) === "[object Date]" && !isNaN(date)) {
-      return new Date(ds[key]).getTime();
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -575,18 +575,18 @@ function sortDataSources(key, order) {
...
-  // Order by updatedAt
   var orderedDataSources = _.orderBy(toBeOrderedDataSources, function(ds) {
-    var date = new Date(ds[key]);
-    if (Object.prototype.toString.call(date) === "[object Date]" && !isNaN(date)) {
-      return new Date(ds[key]).getTime();
+    switch (key) {
+      case 'updatedAt':
+        return new Date(ds[key]).getTime();
+      case 'name':
+        var tempData = ds[key].toUpperCase();
```
squallstar, line 583: `temp` is not allowed as a word, please use a variable name which tells the reader what this is meant to do.

YaroslavOvdii: @squallstar Fixed

PR #83, opened 2020-05-27
[Feat] UI to configure Data Source ACL

squallstar:
ref https://docs.google.com/document/d/1tHDuGBC-zAQOAMXN-rXV2IHELVfWCUTg-s4B-fqvxoo/edit

---

Video ref https://drive.google.com/file/d/1TeDx_NCS940OsUqvndMV67VFVl24uUqU/view?usp=sharing (it's poor quality, I'll retake one more tomorrow)


css/buttons.css
```diff
@@ -0,0 +1,80 @@
+.btn {
```
squallstar, line 1: this is converted from scss, hence why the indentation

tonytlwu: Does the SCSS conversion process not have any configuration or are there no linting tools that can set/fix this to our usual standard?

If not, it's 80 lines of a file. Manually do it if you have to.

templates/accessRule.interface.hbs
```diff
@@ -0,0 +1,10 @@
+<tr data-rule-index="{{index}}">
+    <td class="align-middle">{{{allow}}}</td>
```
tonytlwu, line 2: I can't remember whether we agreed on this, but I remember preferring to align this to the top when there are multiple lines. I understand a lot (not necessarily most) of scenarios would only use one line, which you mentioned. Can you add a screenshot for what it looks like when it's top aligned?

interface.html
```diff
@@ -71,14 +71,17 @@ <h3>No results found</h3>
...
+              <a href="#access-rules" id="show-access-rules" aria-controls="access-rules" role="tab" data-toggle="tab">Security rules</a>
             </li>
             <li role="presentation">
               <a href="#settings" id="show-settings" aria-controls="settings" role="tab" data-toggle="tab">Settings</a>
             </li>
             <li role="presentation" class="hidden">
               <a href="#versions" id="show-versions" aria-controls="versions" role="tab" data-toggle="tab">Version history</a>
             </li>
+            <li role="presentation">
+              <a href="#roles" id="show-users" aria-controls="roles" role="tab" data-toggle="tab">Studio users permissions</a>
```
tonytlwu, line 83:
```suggestion
              <a href="#roles" id="show-users" aria-controls="roles" role="tab" data-toggle="tab">Studio user permissions</a>
```

interface.html
```diff
@@ -213,9 +216,125 @@ <h3>No results found</h3>
...
+                <div class="filters"></div>
+                <button data-add-user-filter type="button" class="btn btn-default"><i class="fa fa-plus fa-fw"></i> Add condition</button>
+              </div>
+            </div>
+          </div>
+          <hr />
+          <div class="row">
+            <div class="col-md-4">
+              <h5>Users can</h5>
+              <p>Select operations the specified users can perform</p>
```
tonytlwu, line 278:
```suggestion
              <p>Select operations the specified users can perform.</p>
```

templates/userMatch.interface.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="required-field">
+  <button class="btn" data-remove-field><i class="fa fa-minus fa-fw fa-lg"></i></button>
+  <input name="column" class="form-control" type="text" placeholder="Name"/>
+  <label class="select-proxy-display">
+    <select class="hidden-select form-control" name="required-field-type">
+      <option value="equals">Equals</option>
+    </select>
+    <span class="icon fa fa-chevron-down"></span>
+  </label>
+  <input name="value" class="form-control" type="text" placeholder=""/>
```
tonytlwu, line 10: Is there a reason you left the placeholder blank?

image

This doesn't exactly say what the last field is for.

```suggestion
  <input name="value" class="form-control" type="text" placeholder="Value"/>
```

squallstar:
@tonytlwu all sorted, see updated screen here:

image

templates/requiredField.interface.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="required-field">
+  <button class="btn" data-remove-field><i class="fa fa-minus fa-fw fa-lg"></i></button>
```
tonytlwu, line 2: I can't compare them side by side using screencasts, but it seems that the minus icon is larger than what's used in the notifications UI. The button is also circular in the latter instead of a pill-shaped button. See https://github.com/Fliplet/fliplet-widget-push-notifications/blob/master/src/components/NotificationForm.vue#L66 for the markup.

Does it resolve both issues if you simply remove the `fa-lg`?

```suggestion
  <button class="btn" data-remove-field><i class="fa fa-minus fa-fw"></i></button>
```

tonytlwu:
I haven't seen any screencasts or screenshots on error checking and validation. Can you show some examples?

squallstar:
@tonytlwu there's only one validation at the moment: a rule can't be saved if you don't select any `type`. The Save button changes state between disabled and active depending on whether you have selected at least one `type`.

What other type of validation and error checking are you looking for?

templates/requiredField.interface.hbs
squallstar: @tonytlwu doesn't look like 
image

tonytlwu: Let's keep it without the `fa-lg` anyway because the + and - icons are much more similar in size this way. If you can, take a look to see why it isn't perfectly circular like the notifications UI, but I'm not that bothered as long as it doesn't use `fa-lg`.

tonytlwu:
@squallstar The way I frame validations is through a list of risk prioritised scenarios that we need to avoid. Some can be automatically avoided through JavaScript and some would need validation to either prevent users from continuing or display warnings:

**High risk** - functionality might partially or completely malfunction

1. Saving with a value that could crash like an invalid Handlebars expression, e.g. `{{ uzzer.[Email address] }}`

**Low risk** - it isn't impossible to decipher why something isn't working, but would be good if the UI took care of it or warned the user ahead of time. If we don't help users avoid them, they need to understand what the error would imply, e.g. saving with **Applies to specific apps** but without any apps means it won't be accessible to any apps.

1. Saving with **Applies to specific apps** but without any apps
1. Saving with **Specific users** but no conditions
1. Saving a **Specific users** or **Data requirement** condition without a field name or uses spaces only
1. Saving a **Specific users** or **Data requirement** condition without a value or uses spaces only (unless a value isn't applicable)
1. Saving without any actions

squallstar:
@tonytlwu ready for review.

PR #85, opened 2020-05-29
[Feat] Refactor to use server-side operators

squallstar:
- Changes to use server-side operators
- Added "not equals" and "contains" in user filter dropdown
- Equals is now case-insensitive

image


tonytlwu:
@squallstar Does the inclusion of sift operator support only give us support for `$ne`? Why was **Not equals** added in reference to this list https://docs.google.com/document/d/1tHDuGBC-zAQOAMXN-rXV2IHELVfWCUTg-s4B-fqvxoo/edit#bookmark=id.kk63kab0qy5z ?

js/interface.js
```diff
@@ -1365,6 +1382,17 @@ $('#show-access-rules').click(function () {
...
 
+  function operatorDescription(operation) {
+    switch(operation) {
+      case 'equals':
+        return 'equals to';
+      case 'notequals':
+        return 'does not equals to';
+      case 'contains':
+        return 'contains';
+    }
```
tonytlwu, line 1393: Let's add a `default` just in case, which should throw an error (because backend won't be able to process it).

js/interface.js
```diff
@@ -1365,6 +1382,17 @@ $('#show-access-rules').click(function () {
   $('.empty-data-source-rules').toggleClass('hidden', currentDataSourceRules.length > 0);
   $('#access-rules-list table').toggleClass('hidden', !currentDataSourceRules.length);
 
+  function operatorDescription(operation) {
+    switch(operation) {
```
tonytlwu, line 1386:
```suggestion
    switch (operation) {
```

js/interface.js
```diff
@@ -1402,7 +1430,16 @@ $('#show-access-rules').click(function () {
             }
 
             return 'Specific users<br />' + _.map(Object.keys(rule.allow.user), function (key) {
-              return '<code>' + key + ' = ' + rule.allow.user[key] + '</code>';
+              var operation = rule.allow.user[key];
+
+              if (typeof operation === 'string') {
```
tonytlwu, line 1435: I've lost track of the overall data structure because we're a few PRs away from the original. Can you explain what this `if` does?

squallstar: this was to support an old syntax, I've removed it.

js/interface.js
tonytlwu: @squallstar I thought about simply returning the operation as its description too. But if we hadn't defined a description, then there's a chance that the operation name isn't supported by the backend. Shouldn't it throw an error instead of letting users think it'll be OK?

squallstar: I don't think we should throw an error. If we ever have other places that can save new rules in the future than the UI should basically show a "non localised label" rather than failing.

tonytlwu: I'm OK with this. FYI if something was suggested, it'll be helpful to provide a reason if the suggestion isn't accepted.

PR #91, opened 2020-06-08
[Maintenance] Post review fixes for acl

squallstar:
### Changes

- Copy update
- Sort apps by name
- Apply changed to have a specific button
- Confirmation message when rules are saved

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -38,7 +38,9 @@ var defaultAccessRules = [
 ];
 
 var getApps = Fliplet.Apps.get().then(function (apps) {
-  return _.sortBy(apps, 'name');
+  return _.sortBy(apps, function (app) {
+    return app.name.toLowerCase();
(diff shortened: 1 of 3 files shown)


js/interface.js
```diff
@@ -37,7 +37,9 @@ var defaultAccessRules = [
   { type: ['delete'], allow: 'all' }
 ];
 
-var getApps = Fliplet.Apps.get();
+var getApps = Fliplet.Apps.get().then(function (apps) {
+  return _.sortBy(apps, 'name');
```
tonytlwu, line 41: To avoid this from happening:

- App
- Zapp
- app
...

```suggestion
  return _.sortBy(apps, function (app) {
    return app.name.toLowerCase();
  });
```

squallstar: thanks, done

PR #101, opened 2020-09-17
[Feature] OD-64 Trash view for data sources

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-64
https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-64
## Description
Created trash view for data source
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYQbl6Jpgm
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


tonytlwu:
@sofiiakvasnevska @upplabs-alex-levchenko Has either of you seen this work? [The screencast](https://screencast-o-matic.com/watch/cYQbl6Jpgm) showed that when the user clicks on Trash, it doesn't appear to be showing the deleted data sources.

sofiiakvasnevska:
@tonytlwu yes, as described in the ticket we have just implemented the view of trash in a current app data 
_"When the user clicks on Trash, it is showing the deleted data sources"_ is described in this ticket
https://weboo.atlassian.net/browse/OD-65 and will be implemented later

tonytlwu:
@sofiiakvasnevska Yes, I'm looking for the ability to view trashed data sources too, not the ability to permanently delete or restore data sources. If you go to 0:09 and 1:05, you'll find that when the user clicks between **All** and **Trash** tabs, they are not shown the deleted data sources.

AndrRyaz:
@tonytlwu, added functionality to trash view, restore and delete

https://screencast-o-matic.com/watch/cYQFDoJKrW

tonytlwu:
- `0:14` When the data source is successfully restored, the data source should be removed from the Trash list instead of continuing to be displayed faded

@squallstar I noticed that when a data source is restored, the **Updated at** timestamp is changed to the time when the data source is restored, and not the last time the data source was updated (before it was deleted). Was this intentional, and if so, why? I'd get confused because the settings and content of the data source have not been updated. If I delete then restore a file in Google Docs, for example, the document doesn't change its last updated timestamp.

squallstar:
@tonytlwu that is intentional and the example you're making in regards to Google Docs is partially different because restoring a file simply restores the file: the entity is "1 to 1" to the content you restore. On the other hand, a data source is "1 to N" to the entries. Restoring entries does not change all settings of the data source, only the entries, but you're essentially making a change to a part of such entity hence the timestamp is updated.

If the timestamp was being restored to the actual version restored, then the timestamp would be different in relation to other changes made to things like settings, hooks and any other change like security which correctly affects the timestamp. Take this example:

5AM Updated data source entries (V1)
6AM Updated data source hooks
7AM Updated data source entries (V2)
8AM Restored data source entries to V1

The timestamp would be shown as 8AM because that's when the data source has changed. If it was changed to 5AM as you suggest then you would miss the information about it being updated at 6AM for different reasons.

tonytlwu:
@squallstar Unless I'm mistaken, it sounds like you were talking about the restoring of a data source version. In that case, I agree with everything you're saying. I'm referring to restoring a data source from trash. The contents of the data source and its entries haven't changed when a data source is restored from trash, so I imagined that the timestamp wouldn't change.

squallstar:
@tonytlwu ah OK I see now, yes in that case I agree the timestamp shouldn't change when the object is restored.

squallstar:
@tonytlwu see https://github.com/Fliplet/fliplet-api/pull/4307 for the fix.

AndrRyaz:
@tonytlwu, @sofiiakvasnevska, added fixes to scenario

https://screencast-o-matic.com/watch/cYQr2DdVrl

css/interface.css
```diff
@@ -12,6 +12,58 @@ body {
...
+
+.btn-round[data-show-alive-source],
+.btn-round[data-show-trash-source] {
+  padding-left: 20px;
+  padding-right: 20px;
+}
+
+.active-source {
+  background-color: #36344c;
+  color: white;
```
tonytlwu, line 33: Use hexcode as much as you can

```suggestion
  color: #fff;
```

css/interface.css
```diff
@@ -24,30 +76,39 @@ body {
   padding-bottom: 120px; /* Compensate for the top part of the Studio UI */
 }
 
-#data-sources {
+#data-sources,
```
tonytlwu, line 79: Give these tables/containers a class instead so you don't need to repeat the CSS selectors

js/interface.js
```diff
@@ -128,14 +132,25 @@ function getDataSources() {
...
   });
 
   $dataSources.html(html.join(''));
   $initialSpinnerLoading.removeClass('animated');
   $contents.removeClass('hidden');
+  $('#trash-sources').hide();
+}
+
+function renderTrashSources(trashSources) {
+  var html = [];
```
tonytlwu, line 146: @upplabs-alex-levchenko Please check with @AndrRyaz on all the usual rules of extra such as empty lines as well as all the linting rules.

```suggestion
  var html = [];

```

js/interface.js
```diff
@@ -128,14 +132,25 @@ function getDataSources() {
...
+  trashSources.forEach(function (trashSource) {
```
tonytlwu, line 147: Use `.map` instead of pushing items into an array with `.forEach`

```suggestion
  trashSources.map(function (trashSource) {
```

js/interface.js
```diff
@@ -595,8 +619,127 @@ function activateFind() {
...
+    });
+
+    Fliplet.Modal.alert({
+      title: 'Restore complete',
+      message: '"' + name + '" restored',
+    });
+
+    renderTrashSources(trashSources);
+  })
+  .catch(function (error) {
```
tonytlwu, line 640: The way `.then()` and `.catch()` are chained to the previous function and indented and so on should be the same

js/interface.js
```diff
@@ -595,8 +619,127 @@ function activateFind() {
...
+    trashSources = trashSources.filter(function(ds) {
+      return ds.id !== id;
+    });
+
+    Fliplet.Modal.alert({
+      title: 'Restore complete',
+      message: '"' + name + '" restored',
+    });
+
+    renderTrashSources(trashSources);
```
tonytlwu, line 638: If the restored data source is removed, why does the list need to be rendered again?

tonytlwu (changes requested):

js/interface.js
AndrRyaz: @tonytlwu can you please explain more about this comment? Do you mean to move then and catch to event function?

tonytlwu: @AndrRyaz It's about consistent point of line break and indentation. @upplabs-alex-levchenko @sofiiakvasnevska Please make sure new developers understand this.

**Don't**

```js
doSomething().then(function () {
  // Function
})
.catch(function () {
  // Function
});

doSomething()
.then(function () {
  // Function
}).catch(function () {
  // Function
});

doSomething()
.then(function () {
  // Function
})
  .catch(function () {
    // Function
  });
```

**Do**

```js
doSomething().then(function () {
  // Function
}).catch(function () {
  // Function
});

doSomething()
  .then(function () {
    // Function
  })
  .catch(function () {
    // Function
  });
```

js/interface.js
tonytlwu: @upplabs-alex-levchenko Please confirm that you have instructed @AndrRyaz about the rules in the use of empty lines and repeat the rules here in a comment so I can confirm it's correct, because I still see many more instances in this PR that violates the rules that other developers seem to have learnt quite successfully.

AndrRyaz:
@tonytlwu resolved all comments about lint fixes, next time I will be more attentive

squallstar:
@AndrRyaz this PR seem to have commits originating from two different github users. One of them isn't in the approved list of contributors for our project. Can you please have a look and confirm?

AndrRyaz:
@squallstar , yes sorry this is mine profile from another computer, forgot to change git user

js/interface.js
tonytlwu: @upplabs-alex-levchenko Please read and respond to the comment above. There are still plenty of errors that other developers seem to have learnt not to make, e.g. mandatory empty lines etc.

squallstar (approved):

PR #110, opened 2020-11-23
OD-250 [Fix] Search will work as expected.

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://weboo.atlassian.net/browse/OD-250

## Description
Search will work as expected.

## Screenshots/screencasts
https://share.getcloudapp.com/NQuKO5bq

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -33,6 +33,8 @@ var table;
 var dataSourceEntriesHasChanged = false;
 var isShowingAll = false;
 var columns;
+var dataSourcesToSearch = [];
+var isShowingTrash = false;
 
 var defaultAccessRules = [
   { type: ['select', 'insert', 'update', 'delete'], allow: 'all' }
@@ -70,7 +72,7 @@ function getDataSources() {
   $('#trash-sources').hide();
 
   return Fliplet.DataSources.get({
-    appId: isShowingAll ? undefined : copyData.appId,
+    appId: isShowingAll ? null : copyData.appId,
     attributes: 'id,name,bundle,createdAt,updatedAt,appId,apps',
     roles: 'publisher,editor',
     type: null
@@ -79,6 +81,7 @@ function getDataSources() {
   })
     .then(function(userDataSources) {
       allDataSources = userDataSources;
+      dataSourcesToSearch = userDataSources;
 
       if (copyData.context === 'app-overlay' || copyData.appId) {
         // Changes UI text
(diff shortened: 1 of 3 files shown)


js/interface.js
```diff
@@ -70,7 +72,7 @@ function getDataSources() {
   $('#trash-sources').hide();
 
   return Fliplet.DataSources.get({
-    appId: isShowingAll ? undefined : copyData.appId,
```
squallstar, line 73: Please explain this change

js/interface.js
```diff
@@ -877,24 +898,23 @@ $('#app')
...
     $('[data-app-source]').addClass('hidden');
     $('[data-show-all-source]').removeClass('hidden');
 
-    if ($('[data-show-trash-source]').hasClass('active-source')) {
-      $('[data-show-trash-source]').click();
-    } else {
-      isShowingAll = false;
+    isShowingAll = false;
 
+    if (isShowingTrash) {
```
squallstar, line 915: couldn't you simply have `getDataSources()` to take care of whether it needs to fetch trashed or non trashed data sources? You literally have the `isShowingTrash` variable for it.

js/interface.js
```diff
@@ -33,6 +33,8 @@ var table;
 var dataSourceEntriesHasChanged = false;
 var isShowingAll = false;
 var columns;
+var dataSourcesToSearch = [];
```
squallstar, line 36: Please explain and make a comment on what `dataSourcesToSearch` is for.

squallstar (changes requested):

js/interface.js
YaroslavOvdii: @squallstar the change was made only to keep consistent, as we are using similar syntax in showTrashedDataSources. As I tested it doesn't change any result from the API. Should we use undefined instead?

js/interface.js
YaroslavOvdii: @squallstar Yes, we could combine getTrashedDataSources and getDataSources but the API that we are using to get trashed and alive data sources return different results and in that case, we will need to use a separate promise chain and it will make the function much larger and harder to read than now.
Should we do that?

js/interface.js
YaroslavOvdii: @squallstar That variable was added in this [PR](https://github.com/Fliplet/fliplet-widget-data-source/pull/68) in order to fix this [issue](https://github.com/Fliplet/fliplet-studio/issues/3841) but there was an error with it because it wasn't declared.

js/interface.js
squallstar: Please use `undefined` if the result is the same

YaroslavOvdii:
The issue was resolved by this [PR](https://github.com/Fliplet/fliplet-widget-data-source/pull/112)

PR #113, opened 2020-12-11
OD-284 [Fix] 'Data source missing rules' when connected to DS provider via API

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-284 https://weboo.atlassian.net/jira/software/projects/OD/boards/9?selectedIssue=OD-284
## Description
Data source provider shows alert 'This data source is missing security rules.' when connecting data source using API
## Screenshots/screencasts
https://screencast-o-matic.com/watch/cYlXbX8fGb
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko


js/interface.js
```diff
@@ -1426,6 +1426,10 @@ $('#show-access-rules').click(function() {
   $tbody.html('');
   $accessRulesList.css('opacity', 0.5);
 
+  if (!Array.isArray(currentDataSourceRules)) {
```
squallstar, line 1429: as I have mentioned previously, if the data sources is not an array it means "default access". It looks like you have coded it to be "no access", which is incorrect. Please explain? Did you read my previous explanation?

AndrRyaz: @squallstar Sorry, yes, I read your comment and I think I understood it the other way around, I will correct the logic and as I understand this change will not only be in the data-source component

AndrRyaz:
@squallstar I will close this PR because the changes here are no longer relevant

PR #117, opened 2021-03-23
OD-351 [Feat] Remove empty columns that have the Column (X) format across the whole data source entries

ivandevupp:
@sofiiakvasnevska
## Issue
OD-351 https://weboo.atlassian.net/browse/OD-351
## Description
Remove empty columns that have the Column (X) format across the whole data source entries on save.
## Screenshots/screencasts
https://monosnap.com/file/BldXrPB5g0jN8yPIk8a7A9yIUSRV2m
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @AndrRyaz @YaroslavOvdii

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -57,7 +57,7 @@ var definitionEditor = CodeMirror.fromTextArea($('#definition')[0], {
   mode: 'javascript'
 });
 
-var emptyColumnNameRegex = /Column\s\([0-9]+\)/;
+var emptyColumnNameRegex = /^Column\s\([0-9]+\)$/;
 
 // Fetch all data sources
 function getDataSources() {
--- a/js/interface.js
+++ b/js/interface.js
@@ -398,13 +398,12 @@ function getEmptyColumns(columns, entries) {
   });
 
   _.forEach(entries, function(entry) {
-    for (var i = 0; i < emptyColumns.length; i++) {
+    for (var i = emptyColumns.length - 1; i >= 0; i--) {
       if (entry.data[emptyColumns[i]] !== null && entry.data[emptyColumns[i]] !== undefined && entry.data[emptyColumns[i]] !== '') {
         var notEmptyColumnIndex = emptyColumns.indexOf(emptyColumns[i]);
   
         if (notEmptyColumnIndex !== -1) {
           emptyColumns.splice(notEmptyColumnIndex, 1);
-          i--;
         }
       }
     }
--- a/js/interface.js
+++ b/js/interface.js
@@ -397,21 +397,24 @@ function getEmptyColumns(columns, entries) {
     return emptyColumnNameRegex.test(column);
   });
 
-  _.forEach(entries, function(entry) {
-    if (!emptyColumns.length) {
-      return false;
-    }
-
-    for (var i = emptyColumns.length - 1; i >= 0; i--) {
-      if (entry.data[emptyColumns[i]] !== null && entry.data[emptyColumns[i]] !== undefined && entry.data[emptyColumns[i]] !== '') {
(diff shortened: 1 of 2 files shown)


js/interface.js
```diff
@@ -390,6 +392,35 @@ function trimColumns(columns) {
   });
 }
 
+function getEmptyColumns(columns, entries) {
+  var emptyColumns = _.filter(columns, function(column) {
+    return emptyColumnNameRegex.test(column);
+  });
+
+  _.forEach(entries, function(entry) {
+    for (var i = emptyColumns.length - 1; i >= 0; i--) {
```
squallstar, line 401: what is this for cycle supposed to do?

ivandevupp: This cycle is used to iterate through `emptyColumns` (columns that have the `Column (X)` format) for each entry and check if column has a value.
When column has value in some entry - then we remove this column from `emptyColumns`.
This cycle is descending to remove items from `emptyColumns` without index manipulations.
Also I wanted to use Handsontable's `isEmptyCol` but it only checks `null` and `undefined` values (not empty strings).

UPD: I added check at the beginning of `forEach` if `emptyColumns` is empty to stop iterating through all entries.

js/interface.js
```diff
@@ -57,6 +57,8 @@ var definitionEditor = CodeMirror.fromTextArea($('#definition')[0], {
   mode: 'javascript'
 });
 
+var emptyColumnNameRegex = /Column\s\([0-9]+\)/;
```
tonytlwu, line 60: It's important that this is the entire column name.

```suggestion
var emptyColumnNameRegex = /^Column\s\([0-9]+\)$/;
```

js/interface.js
```diff
@@ -390,6 +392,39 @@ function trimColumns(columns) {
   });
 }
 
+function getEmptyColumns(columns, entries) {
+  var emptyColumns = _.filter(columns, function(column) {
+    return emptyColumnNameRegex.test(column);
+  });
+
+  _.forEach(entries, function(entry) {
+    if (!emptyColumns.length) {
```
tonytlwu, line 401: You'll want this outside the _.forEach loop too because if there is no empty column, there's no need to go through the _.forEach loop at all. The current code could end up going through this `if` check `entries.length` number of times.

tonytlwu (changes requested):

js/interface.js
ivandevupp: Fixed.

js/interface.js
ivandevupp: I added condition outside the loop and added comment for this row. We still need to check `emptyColumns.length` inside _.forEach loop because when all `emptyColumns` have values on `N` iteration (`N < entries.length`) we must stop iterating through entries.

tonytlwu: Yes, keeping it inside the loop makes sense too.

js/interface.js
```diff
@@ -390,6 +392,42 @@ function trimColumns(columns) {
   });
 }
 
+function getEmptyColumns(columns, entries) {
+  var emptyColumns = _.filter(columns, function(column) {
+    return emptyColumnNameRegex.test(column);
+  });
+
+  if (emptyColumns.length) {
```
tonytlwu, line 400: Avoid the `if` indentation with this.

```suggestion
  if (!emptyColumns.length) {
    return emptyColumns;
  }
```

tonytlwu (changes requested):

js/interface.js
ivandevupp: Fixed.

tonytlwu (approved):
@sofiiakvasnevska This is ready for testing

Risk: Critical
Affected: Data source manager users in Studio
Verify: Data isn't lost when saving changes

PR #125, opened 2021-04-01
OD-459 [Fix] Ensure sorting works after sorting multiple times

AndrRyaz:
@sofiiakvasnevska 
## Issue
OD-459 https://weboo.atlassian.net/browse/OD-459
## Description
Added fixes to sorting in data source manager
## Screenshots/screencasts
http://somup.com/crfVQfb5hD
## Backward compatibility
## Reviewers
@upplabs-alex-levchenko @ivandevupp @YaroslavOvdii

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -874,150 +874,58 @@ $(window).on('resize', windowResized).trigger('resize');
 $('#app')
   .on('click', '[data-order-date]', function() {
     var $dataSource = $(this);
-    var orderedDataSources;
+    var isOrderedByDateAsc = $dataSource.hasClass('asc');
+    var newOrder = isOrderedByDateAsc ? 'desc' : 'asc';
+    var orderedDataSources = sortDataSources('updatedAt', newOrder, dataSources);
 
-    if ($dataSource.hasClass('desc')) {
-      $dataSource.removeClass('desc').addClass('asc');
-      // Order data sources by updatedAt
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -874,150 +874,58 @@ $(window).on('resize', windowResized).trigger('resize');
 $('#app')
   .on('click', '[data-order-date]', function() {
     var $dataSource = $(this);
-    var orderedDataSources;
+    var isOrderedByDateAsc = $dataSource.hasClass('asc');
```
tonytlwu, line 877: I understand the logic here but we've got the same logic applied 5 times. Can this be modularised to keep it DRY?

tonytlwu (changes requested):

js/interface.js
AndrRyaz: @tonytlwu i added a new function to make it more clear

tonytlwu (approved):
@sofiiakvasnevska This is ready for testing

Risk: Medium
Affected: Data source manager users sorting data sources
Verify: Data sources can be sorted as expected

@AndrRyaz The PR title is too vague.

`Changes to sorting in data source manager` is only as good as `Changes to logic`. You need to specify what kind of changes or what kind of benefit it will bring to users e.g. `Ensure sorting works after sorting multiple times`

PR #129, opened 2021-04-02
OD-450 [Fix] Parse nested arrays and objects for render in spreadsheet correctly

ivandevupp:
@sofiiakvasnevska
## Issue
OD-450 https://weboo.atlassian.net/browse/OD-450
## Description
Parse nested arrays and objects for render in spreadsheet correctly
## Screenshots/screencasts
https://monosnap.com/file/MdPV4ocMYAKreLFsf8fdEEEqRmJPFr
## Backward compatibility
## Reviewers 
@upplabs-alex-levchenko @AndrRyaz @YaroslavOvdii

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -49,7 +49,7 @@ var spreadsheet = function(options) {
               return JSON.stringify(val);
             }
 
-              return typeof val === 'string' && val.indexOf(',') !== -1 ? '"' + val + '"' : val;
+            return typeof val === 'string' && val.indexOf(',') !== -1 ? '"' + val + '"' : validateJsonString(val);
           }).join(', ');
         // Stringify value only for the first render for nested objects
         } else if (isFirstRender && value && typeof value === 'object') {
@@ -66,6 +66,16 @@ var spreadsheet = function(options) {
     preparedData.unshift(columns);
 
     return preparedData;
+  }
+
+  function validateJsonString(str) {
+    var validatedString;
+    try {
+      validatedString = JSON.parse(str);
+    } catch (e) {
+      validatedString = str;
+    }
+    return validatedString;
   }
 
   function onChanges() {
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -8,6 +8,8 @@ var data;
 var colWidths = [];
 var s = [1, 0, 1, 0]; // Stores current selection to use for toolbar
 
+var jsonObjRegExp = /^({|\[).*(}|\])$/;
+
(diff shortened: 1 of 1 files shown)


squallstar:
Can you confirm the data is saved back to json when read via the APIs?

ivandevupp:
> Can you confirm the data is saved back to json before saving? Please check that when reading after saving the data is actually a JSON rather than a JSON string.

@squallstar Because of `prepareData` fires on first table render and in `afterChangesObserved` hook, data is stringified only on render and it is not saved back as a JSON string, because users can save datasource only after changes are made, but if changes are made we don't change data in `prepareData` function.

squallstar:
@ivandevupp does it mean that if I make a change to the string then it's not converted back to a JSON? If so, that's basically breaking the functionality we want to give users.

They should be able to amend the JSON and such JSON should correctly be saved as a JSON (not a JSON string) when the record is updated. @tonytlwu can you confirm?

ivandevupp:
> @ivandevupp does it mean that if I make a change to the string then it's not converted back to a JSON? If so, that's basically breaking the functionality we want to give users.
> 
> They should be able to amend the JSON and such JSON should correctly be saved as a JSON (not a JSON string) when the record is updated. @tonytlwu can you confirm?

@squallstar I added functionality that allows users to amend and correctly save the JSON values.
WHEN we add entries with JSON props to the datasource, they are added and rendered correctly:
image
image

WHEN we change some JSON value and commit it to the datasource:
image

THEN after page reload we have correct data in query:
image

squallstar:
@ivandevupp it looks like you're parsing all inputs into JSON, which will result in some strings being casted to a Number or Boolean or else (e.g. `JSON.parse("false")` is `false`) and we don't want that. Please only parse JSON strings that are either objects or arrays, e.g. they start either with `{` or `[`

ivandevupp:
@squallstar Update with parsing only JSON strings that are either objects or arrays by regexp: `^({|\[).*(}|\])$`

squallstar:
@ivandevupp can you update the regexp to allow empty spaces before and after the json object? e.g. `[ ]*`

ivandevupp:
@squallstar you mean `^[\s]*({|\[).*(}|\])[\s]*$` ?

squallstar:
@ivandevupp yes, that works.

squallstar (approved):
@sofiiakvasnevska this is ready to be deployed on **staging** for QA.

- **Risk**: Critical
- **Affected**: Studio users managing data sources
- **Verify**: Verify data in data sources (including JSON data) can be managed

PR #138, opened 2021-05-11
ID-1026 [Maintenance] Speeds up initialization of Handsontable

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1026

- Speeds up computation of unique column names by 4~5x https://jsbench.me/p3koki07ry/2
- Reduces Handsontable initialisation time by ~33% by avoiding a re-render right after initialisation
- Reduces time spent in custom render functions by ~50% using a more efficient wrapping function
- More accurately use column header widths
- On initial render, loads an empty Handsontable before loading data. This speeds up load time from ~160 seconds to ~5 seconds for 10k entries.

## Results

Includes changes from https://github.com/Fliplet/fliplet-widget-data-source/pull/137/commits/7072b6477915e173c4b059784a4840577238384b from [this PR](https://github.com/Fliplet/fliplet-widget-data-source/pull/137)

### Rendering 200 entries

<details>
<summary>Before (~5 seconds)</summary>

image
</details>

<details>
<summary>After (~2 seconds)</summary>

image
</details>

### Rendering 10k entries

<details>
<summary>Before (~160 seconds)</summary>

Profiler tool wasn't able to profile for more than 1.5 minutes.

image

</details>

<details>
<summary>After (~6 seconds)</summary>

image

</details>

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -35,6 +35,7 @@ var dataSourceEntriesHasChanged = false;
 var isShowingAll = false;
 var columns;
 var dataSourcesToSearch = [];
+var initialLoad = true;
 
 var defaultAccessRules = [
   { type: ['select', 'insert', 'update', 'delete'], allow: 'all' }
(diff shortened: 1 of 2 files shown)


squallstar (approved):
Wow, this is quite a significant improvement! Do you think that performances are now to a state where we don't need to discuss changing the core library to move away from Handsontable?

tonytlwu:
@squallstar Assuming a few other performance issues are resolved too, potentially. Once we get data saving to be performant too, then it could be a lot easier to use. I've only tested up to 10k entries and we already have apps that are going up towards 100k entries within a year's time.

At some point pagination would be the answer, and we'd also need column types too. At that point, we may still need to assess whether the paradigm needs to change from a spreadsheet model to a database management (row-by-row) model -- to determine if a change in core library would be suitable.

PR #152, opened 2021-06-21
ID-1161 [Chore] Updates copy for excluded columns following backend updates

squallstar:
> Specify columns that should never be readable (or writeable, depending on the operation) to users when this rule is matched.

### DIFF
--- a/interface.html
+++ b/interface.html
@@ -358,7 +358,7 @@
           <div class="row">
             <div class="col-md-4">
               <h5>Columns to exclude</h5>
-              <p>Specify columns that should never be readable (or writeable, depending on the operation) to users when this rule is matched.</p>
+              <p data-exclude-description></p>
             </div>
             <div class="col-md-8 side">
               <div class="tokenfield-ui">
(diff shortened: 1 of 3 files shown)


interface.html
```diff
@@ -358,7 +358,7 @@ <h5>Applies to</h5>
           <div class="row">
             <div class="col-md-4">
               <h5>Columns to exclude</h5>
-              <p>Specify columns that should never be returned to users when this rule is matched.</p>
+              <p>Specify columns that should never be readable (or writeable, depending on the operation) to users when this rule is matched.</p>
```
tonytlwu, line 361: **Writable** is more commonly used than **writeable**.

Also, I think we should just show "readable to" or "writable by" depending on the type of rule, rather than displaying a complex sentence for users to decode.

squallstar: @tonytlwu a rule can affect both read and write at the same time though, what do we do in that case?

tonytlwu: Adapt according to the selected rules?

- readable by
- readable or writable by
- writable by

I'm changing **readable to** to **readable by** here. Both make sense to me and is easier. I just don't want to show "readable or writable by" to the user when the rule doesn't affect writing. Users could get confused here, even if we said "depending on the rule".

tonytlwu (approved):

tonytlwu (approved):

PR #154, opened 2021-06-25
OD-534 [FIX] Force update of the Data Source security rules

S-Volodymyr:
DS security rule UI does not update correctly


squallstar (approved):
@sofiiakvasnevska this is ready to be deployed on **staging** for QA.

- **Risk**: Low
- **Affected**: Studio users configuring data source access rules with exclusion fields
- **Verify**: Verify the issue is resolved

PR #160, opened 2021-07-30
OD-508 [FIX] Fixed to saving size after moving columns

OleksandrLavrynenko:
@sofiiakvasnevska @inna-bieshulia

OD-508 https://weboo.atlassian.net/browse/OD-508

## Description
Fixed to saving size after moving columns. The column is now resized after being moved.

## Screenshots/screencasts
https://storyxpress.co/video/krqjvs9ehxp7ssjwe

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko @S-Volodymyr @skrupnyk

### DIFF
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -406,12 +406,11 @@ var spreadsheet = function(options) {
     sortIndicator: true,
     selectionMode: 'range',
     cells: function(row) {
-      /* eslint-disable consistent-return */
       if (row !== 0) {
-        return;
+        return; // eslint-disable-line consistent-return
       }
 
-      return {
+      return { // eslint-disable-line consistent-return
         renderer: columnValueRenderer
       };
     },
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -407,10 +407,10 @@ var spreadsheet = function(options) {
     selectionMode: 'range',
     cells: function(row) {
       if (row !== 0) {
+        // eslint-disable-next-line
         return;
       }
 
-      // eslint-disable-next-line
       return {
         renderer: columnValueRenderer
       };
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -342,7 +342,7 @@ var spreadsheet = function(options) {
   /**
    * Style the first row (columns headings)
    */
-  function columnValueRenderer(instance, td, row, col, prop, value, cellProperties) {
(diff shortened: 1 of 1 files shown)


js/spreadsheet.js
```diff
@@ -531,6 +531,26 @@ var spreadsheet = function(options) {
 
       colWidths = getColWidths();
     },
+    beforeColumnMove: function(items, index) {
+      colWidths = getColWidths();
+
+      var length = items.length;
+      var colsToMove = colWidths.splice(items[0], length);
```
squallstar, line 538: any chance `items` can be empty here?

OleksandrLavrynenko: > any chance `items` can be empty here?

the hook "beforeColumnMove" is activated when at least one item is selected

squallstar (approved):
@inna-bieshulia this is ready to be deployed on **staging** for QA.

- **Risk**: Medium
- **Affected**: Studio users using the data source manager
- **Verify**: Verify the issue is resolved

OleksandrLavrynenko:
@tonytlwu could you please advice what we need to do with the function "cells"? It is an old code and Lint does not allow it. Should we rewrite it or add ignore?

tonytlwu:
@OleksandrLavrynenko What do you mean by the Lint doesn't allow it? What's rule is it breaking and what's the error?

upplabs-alex-levchenko:
@tonytlwu It's about Codacy Static Code Analysis, it says - "A function should not mix return statements with and without a result".

tonytlwu:
@upplabs-alex-levchenko Thanks for the clarification. I'd ignore it by adding `// eslint-disable-next-line consistent-return`. I'm not sure which line to add it in (before the `function` definition or the `return` statement). I'll let you guys figure that out.

upplabs-alex-levchenko:
Thanks, roger that.

upplabs-alex-levchenko:
@tonytlwu we've tried different annotations and their positions but it does not affect Codacy for some reason. Please ignore this error through their UI cuz it's not available for us. Thanks. 
image

tonytlwu:
@upplabs-alex-levchenko Let's keep this moving forward as it is then. We don't want to globally disable the rule because it's a good sense check for inconsistencies, e.g. some functions need to always return a Promise and it would be bad if one of the clauses returned `undefined`.

@inna-bieshulia You can merge this and continue with testing now.

PR #166, opened 2021-10-27
OD-602 [Fix] Fixed bug in editing objects and arrays

OleksandrLavrynenko:
@romanyosyfiv

OD-602 https://weboo.atlassian.net/browse/OD-602

## Description
**Problem**: The Papa.js library corrupted  objects and arrays in JSON. This made it impossible to parse the data from JSON.
**Solution**: Added parsing function Objects and arrays in JSON. This function checks for an object or array in JSON and unpacks them.

## Screenshots/screencasts

https://user-images.githubusercontent.com/86256215/141954154-1500186c-6b63-48c9-be26-20466b2f1306.mp4

## Backward compatibility

## Reviewers
@upplabs-alex-levchenko


js/spreadsheet.js
```diff
@@ -861,16 +840,32 @@ var spreadsheet = function(options) {
     return entries;
   };
 
-  function validateJsonString(str) {
-    var validatedString;
-
+  function getColumnValue(str) {
     try {
-      validatedString = jsonObjRegExp.test(str) ? JSON.parse(str) : str;
+      var parsingResult = JSON.parse('[' + str + ']');
```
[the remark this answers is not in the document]
tonytlwu: @OleksandrLavrynenko Based on what you're saying, an array of objects are joined using `,` and printed on Handsontable, and that's why [{foo: 1}] --> '{"foo": 1}'. Is that right?

If so, then it sounds like it shouldn't be displayed by joining with `,`. If we have the right capability to parse and display JSON, then it can just be stringified before displaying with Handsontable, and parsed into JSON before it's saved.

Would that solve the problem?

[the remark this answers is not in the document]
OleksandrLavrynenko: **@tonytlwu** I converted arrays and objects to JSON for display. Are you satisfied with this format for display arrays and objects?
obj-arr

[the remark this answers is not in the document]
tonytlwu: @OleksandrLavrynenko This looks good. How does it deal with other types, e.g. string, integers, booleans.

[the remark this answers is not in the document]
OleksandrLavrynenko: **@tonytlwu** strings, numbers and booleans are converted to a string after editing or at the moment of entering them into the table, just as it works now.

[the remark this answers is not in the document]
tonytlwu: @OleksandrLavrynenko Can you show screenshots with the examples of the various types just as you did in your previous comments please?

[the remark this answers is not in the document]
OleksandrLavrynenko: **@tonytlwu** Here is a screenshot of the sent data and its output in the table
types
arr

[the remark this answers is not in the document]
tonytlwu: @OleksandrLavrynenko Looks good to me. Please continue with the solution.

js/spreadsheet.js
```diff
@@ -36,31 +33,30 @@ var spreadsheet = function(options) {
...
-            return typeof val === 'string' && val.indexOf(',') !== -1 ? '"' + val + '"' : val;
-          }).join(', ');
         // Stringify value only for the first render for nested objects
-        } else if (isFirstRender && value && typeof value === 'object') {
+        if (isFirstRender && typeof value === 'object') {
           if (objColumns.indexOf(header) === -1) {
             objColumns.push(header);
           }
 
+          if (Array.isArray(value)) {
```
tonytlwu, line 48: @OleksandrLavrynenko I'm reading this section and there are a few things that don't make sense to me.

**Objective**

- Stringify objects/arrays so that they are representative of the content, e.g. arrays of objects use `[]` as the wrapper as expected, objects aren't displayed as `[object Object]` etc.
- Objects/arrays are saved correctly

I therefore see a few problems here:

1. Why do we need the logic to add double quotes to the string if it contains a comma?
2. Why do we have a special scenario for nested arrays?

Could you also share screenshots of what this looks like if the data was the following:

* `[1, "2, and 3", 4, { foo: "bar" }]`
* `[1, 2, [3, 4, "hi, how are you"], 5, 6, { foo: ["bar", "buzz"] }]

OleksandrLavrynenko: **@tonytlwu**  There is a problem with parsing the first line

compare-dsta

tonytlwu: @OleksandrLavrynenko I also don't see `[1, 2, [3, 4, "hi, how are you"], 5, 6, { foo: ["bar", "buzz"] }]` being rendered correctly. Can you fix both issues please?

I expect the following:

* `[1, "2, and 3", 4, { foo: "bar" }]` => `[1,"2, and 3",4,{"foo":"bar"}]`
* `[1, 2, [3, 4, "hi, how are you"], 5, 6, { foo: ["bar", "buzz"] }]` => `[1,2,[3,4,"hi, how are you"],5,6,{"foo":["bar","buzz"]}]`

js/spreadsheet.js
```diff
@@ -36,32 +33,31 @@ var spreadsheet = function(options) {
...
         // Stringify value only for the first render for nested objects
-        } else if (isFirstRender && value && typeof value === 'object') {
+        if (isFirstRender && typeof value === 'object') {
           if (objColumns.indexOf(header) === -1) {
             objColumns.push(header);
           }
 
-          value = JSON.stringify(value);
+          if (Array.isArray(value)) {
+            // Add double quotes to the string if it contains a comma
```
tonytlwu, line 49: Is this comment still relevant? It doesn't look like we're adding double quotes any more.

Please review all the inline comments in this function to make sure they are relevant and correct.

js/spreadsheet.js
```diff
@@ -36,32 +33,31 @@ var spreadsheet = function(options) {
...
-          // Add double quotes to the string if it contains a comma
-          value = value.map(function(val) {
-            // Stringify value only for the first render for nested arrays
-            if (isFirstRender && value && typeof val !== 'string') {
-              return JSON.stringify(val);
-            }
-
-            return typeof val === 'string' && val.indexOf(',') !== -1 ? '"' + val + '"' : val;
-          }).join(', ');
         // Stringify value only for the first render for nested objects
```
tonytlwu, line 42: What's the scenario that this comment is describing?

js/spreadsheet.js
tonytlwu: @OleksandrLavrynenko Please show screenshot.

tonytlwu (changes requested):

js/spreadsheet.js
```diff
@@ -36,28 +33,15 @@ var spreadsheet = function(options) {
...
-            return typeof val === 'string' && val.indexOf(',') !== -1 ? '"' + val + '"' : val;
-          }).join(', ');
-        // Stringify value only for the first render for nested objects
-        } else if (isFirstRender && value && typeof value === 'object') {
-          if (objColumns.indexOf(header) === -1) {
+        // Stringify values for rendering objects for the first rendering and when undoing changes to the table (Ctrl + Z in the table cell)
+        if (typeof value === 'object' && value !== null) {
+          if (isFirstRender && objColumns.indexOf(header) === -1) {
             objColumns.push(header);
           }
```
tonytlwu, line 47: The next line `value = JSON.stringify(value);` used to only execute if `isFirstRender` is true. Now it's executed even if it isn't `true`. Is that intentional and if so, why is it necessary?

js/spreadsheet.js
```diff
@@ -861,16 +825,32 @@ var spreadsheet = function(options) {
...
+      return getString(str);
     } catch (e) {
-      validatedString = str;
+      return getString(str);
     }
+  }
 
-    return validatedString;
+  function getString(str) {
+    try {
```
tonytlwu, line 843: Why is this part necessary? What's the problem scenario if we don't use this in the `getColumnValue` function?

js/spreadsheet.js
OleksandrLavrynenko: I found another bug in the table. I checked, it's reproduced on the staging. 

1. Edit the object.
2. Select another cell to save changes to the object.
3. Select the cell from the object again and press backspace
4. Press Ctrl + Z to return the object.
All objects in the table will become [Object object]

tonytlwu: @OleksandrLavrynenko 👍🏻  Can you add a comment in the Jira ticket and mention this bug please, including the steps to replicate.

cc @inna-bieshulia

tonytlwu (approved):
@romanyosyfiv @inna-bieshulia This is ready for testing

Risk: Critical
Affected: Studio users managing data source data
Verify: Data source data is consistently saved, loaded and no data is lost

PR #172, opened 2022-03-10
ID-1986 [Maintenance] Adds ESLint

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1986

This is preparation work for adding selective commits to DS Manager.


squallstar (approved):

js/spreadsheet.js
```diff
@@ -1111,6 +1120,32 @@ function undoRedoToggle() {
   $('[data-action="redo"]').prop('disabled', disableRedo);
 }
 
+function undoRedo(event) {
```
squallstar, line 1123: any specific reason you moved this in here?

tonytlwu: This file already contained some logic and functions related to other undo/redo functions so it was either move them all the the `undo-redo` file, or start adding those function names as global variable. It's all quite messy so I just put them all in the same file.

PR #173, opened 2022-03-15
ID-1028 [Feat] Selectively commit changes to the data source

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1028


js/history.js
```diff
@@ -0,0 +1,126 @@
+Fliplet.Registry.set('history-stack', (function() {
```
squallstar, line 1: What does this library do? May be worth adding a comment

js/interface.js
```diff
@@ -33,11 +33,15 @@ var dataSources;
...
-var dataSourceEntriesHasChanged = false;
 var isShowingAll = false;
 var columns;
 var dataSourcesToSearch = [];
 var initialLoad = true;
+var entryMap = {
+  original: [],
+  entries: []
+};
+var locale = navigator.language.indexOf('en') === 0 ? navigator.language : 'en';
```
squallstar, line 44: What's this for?

js/interface.js
```diff
@@ -567,12 +655,26 @@ function saveCurrentData() {
     return Fliplet.DataSources.update(currentDataSourceId, { definition: dataSource.definition });
   }).catch(console.error);
 
-  currentDataSourceUpdatedAt = moment().format('MMM Do YYYY, HH:mm');
+  currentDataSourceUpdatedAt = TD(new Date(), { format: 'lll', locale: locale });
+
+  var replace = false; // Set to TRUE to commit the changes by replacing all existing entries
```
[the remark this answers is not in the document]
tonytlwu: @squallstar Yes, potentially. I'm currently waiting for your assessment on the commit endpoint to determine what's the best way of using this. For example, if all the rows are to be updated, it could be faster to set `replace = true` to use a different payload. It hasn't been decided yet.

js/interface.js
tonytlwu: This is to make sure any date localisation uses an English locale that's closest to the browser language.

js/history.js
tonytlwu: Added

squallstar (approved):

PR #196, opened 2022-09-30
Removes ES6 syntax to reactivate ESLint

tonytlwu:
The ES6 syntax is preventing ESLint from working as it's catching a series of syntax errors that aren't supported by ES5.

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -301,6 +301,7 @@ function fetchCurrentDataSourceUsers() {
  * Cache a list of entries as original entries for comparison when committing changes
  * @param {Array} entries - Entries to be cached as original entries
  * @param {Object} [clientIdMap] - Optional map of client IDs to new entry IDs to map add the missing entry IDs. This mutates the entries provided.
+ * @returns {undefined}
  */
 function cacheOriginalEntries(entries, clientIdMap) {
   entryMap.original = {};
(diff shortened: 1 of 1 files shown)


js/interface.js
```diff
@@ -359,7 +360,11 @@ function fetchCurrentDataSourceEntries(entries) {
     } else {
       var flattenedColumns = {};
 
-      rows.map(({ data }) => data).forEach(dataItem => (flattenedColumns = { ...flattenedColumns, ...dataItem }));
+      rows.map(function(row) {
```
tonytlwu, line 363: @squallstar Can you confirm if this is still correctly optimised based on the work done in https://github.com/Fliplet/fliplet-widget-data-source/pull/177

squallstar (approved):

js/interface.js
squallstar: @tonytlwu is there a specific reason you want to change the current functionality not to use the spread operator?

Your solution seems faster but creates almost twice as many operations?

tonytlwu: @squallstar Because it isn't allowed for ES5 and our ESLint has not been updated to support this.

tonytlwu: I think high ops/s is meant to be a good thing though.

squallstar (approved):

PR #198, opened 2022-10-07
XOR-563[FIX] Ensure create record automation copies records in the correct order

SMohol:
### Product areas affected

- `interface.js` 

### List the product areas affected
 - Fliplet Widget - fliplet-widget-data-source ` interface.js`

### What does this PR do?

- This PR changed to specify the sort order as order ASC, id ASC.
- Accept input in the request body as follow:
```
     {
        order: [
          ['order', 'ASC'],
          ['id', 'ASC']
        ]
      }
```
### JIRA ticket

- [XOR-563](https://weboo.atlassian.net/browse/XOR-563)

### Result
- Video recording 

https://user-images.githubusercontent.com/107605720/194559633-3fb452d3-ccf3-4713-bff5-96da784da4c3.mp4

https://user-images.githubusercontent.com/107605720/194559659-7b19c098-a133-4c28-9309-7891383da5a2.mp4

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -370,7 +370,7 @@ function fetchCurrentDataSourceEntries(entries) {
       }
 
       return source.find({
-        order: [
+        orderBy: [
           ['order', 'ASC'],
           ['id', 'ASC']
         ]
--- a/js/interface.js
+++ b/js/interface.js
@@ -369,7 +369,12 @@ function fetchCurrentDataSourceEntries(entries) {
         return Promise.resolve(entries);
       }
 
-      return source.find({}).catch(function() {
+      return source.find({
+        orderBy: [
+          ['order', 'ASC'],
+          ['id', 'ASC']
+        ]
+      }).catch(function() {
         return Promise.reject('Access denied. Please review your security settings if you want to access this data source.');
       });
     });
--- a/js/interface.js
+++ b/js/interface.js
@@ -370,7 +370,7 @@ function fetchCurrentDataSourceEntries(entries) {
       }
 
       return source.find({
-        orderBy: [
+        order: [
           ['order', 'ASC'],
           ['id', 'ASC']
         ]


js/interface.js
```diff
@@ -369,7 +369,12 @@ function fetchCurrentDataSourceEntries(entries) {
         return Promise.resolve(entries);
       }
 
-      return source.find({}).catch(function() {
+      return source.find({
+        orderBy: [
```
squallstar, line 373: As per tech spec (https://weboo.atlassian.net/browse/XOR-563?focusedCommentId=36803), the column should be called `order`, not `orderBy`.

squallstar (changes requested):

js/interface.js
SMohol: In the existing endpoint `order` name is already taken so added `orderBy` [here](https://github.com/Fliplet/fliplet-api/blob/%40Smoholxor/FIX-XOR-563/routes/v1/data-sources.js#:~:text=%5B%27id%27%2C%20%27createdAt%27%2C%20%27updatedAt%27%2C%20%27order%27%5D.forEach((key)%20%3D%3E%20%7B)

SMohol: yes we supposed to add the same however body is already having `order `field e.g here(https://github.com/Fliplet/fliplet-api/blob/%40Smoholxor/FIX-XOR-563/routes/v1/data-sources.js#:~:text=%5B%27id%27%2C%20%27createdAt%27%2C%20%27updatedAt%27%2C%20%27order%27%5D.forEach((key)%20%3D%3E%20%7B) so to avoid conflict with name added `orderBy` instead of `order`

SMohol: @squallstar as per https://github.com/Fliplet/fliplet-api/pull/5528#discussion_r990591664 comment  I do changes accordingly please review

squallstar (approved):

PR #200, opened 2022-10-17
XOR-543 [FIX] write/update data in the columns that are excluded by a DS security rule issue fixed.

SMohol:
### Product areas affected
- Datasource write/update operation on data restriction the basis of DS security rule.

### What does this PR do?

- Spilt exclude columns list with ', ' as it contains space between columns names pass from UI for exclude columns

### JIRA ticket
[XOR-543](https://weboo.atlassian.net/browse/XOR-543)

### Result

Relevant sample input/output, screenshots or screen recordings.

https://user-images.githubusercontent.com/107605720/196167191-941905e6-563e-4a43-ae79-05281c019f86.mp4

https://user-images.githubusercontent.com/107605720/196167283-cff29b6f-3a10-4862-a49d-06dbf3f7cb42.mp4

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -2538,7 +2538,7 @@ $('[data-save-rule]').click(function(event) {
   }
 
   // Spilt exclude columns list with ', ' as it contain space between columns names
-  var exclude = _.compact($('input[name="exclude"]').val().split(', '));
+  var exclude = _.compact($('input[name="exclude"]').val().split(',').map(s=>s.trim()));
 
   if (columnsListMode === 'exclude') {
     if (exclude.length) {
--- a/js/interface.js
+++ b/js/interface.js
@@ -2538,7 +2538,7 @@ $('[data-save-rule]').click(function(event) {
   }
 
   // Spilt exclude columns list with ', ' as it contain space between columns names
-  var exclude = _.compact($('input[name="exclude"]').val().split(',').map(s=>s.trim()));
+  var exclude = _.compact($('input[name="exclude"]').val().split(',').map(column => column.trim()));
 
   if (columnsListMode === 'exclude') {
     if (exclude.length) {
--- a/js/interface.js
+++ b/js/interface.js
@@ -2537,7 +2537,6 @@ $('[data-save-rule]').click(function(event) {
     rule.require = requiredFields;
   }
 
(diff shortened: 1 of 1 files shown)


amolsuryawanshi47 (approved):

js/interface.js
```diff
@@ -2537,7 +2537,8 @@ $('[data-save-rule]').click(function(event) {
     rule.require = requiredFields;
   }
 
-  var exclude = _.compact($('input[name="exclude"]').val().split(','));
+  // Spilt exclude columns list with ', ' as it contain space between columns names
```
squallstar, line 2540: 1. Typo in "Split"
2. I suggest using functional programming to split by `,` and then `trim` the results

```js
array.split(',').map(s => s.trim())
```

cc @tonytlwu

squallstar (changes requested):

js/interface.js
SMohol: @squallstar changes added in next commit thank you for inputs

amolsuryawanshi47 (approved):
Changes looks ok to me

js/interface.js
```diff
@@ -2537,7 +2537,8 @@ $('[data-save-rule]').click(function(event) {
...
+  var exclude = _.compact($('input[name="exclude"]').val().split(',').map(s=>s.trim()));
```
squallstar, line 2541: Doesn't seem to be in line with ESLint. I am expecting to see `s => s.trim()`

Can you also name `s` accordingly to its context? I would suggest `column`

squallstar (changes requested):

js/interface.js
SMohol: ok i will check and add accordingly

js/interface.js
squallstar: @SMohol there's still a typo in your comment, but let's please just remove the comment entirely as it doesn't make sense anymore.

squallstar (changes requested):

js/interface.js
SMohol: while testing this column list i got ' ,  ' instead of ',' that is why added a comment accordingly to rectify but removing it now.

squallstar (approved):

PR #210, opened 2023-01-18
ID-2830 [Chore] Fetch api tokens from all apps to populate tokens dropdown

squallstar:
image

requires https://github.com/Fliplet/fliplet-api/pull/5629

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -43,7 +43,7 @@ var entryMap = {
   entries: {}
 };
 var currentFinalRules;
-var integrationTokenList;
+var integrationTokenList = [];
 var selectedTokenId;
 var selectedTokenName;
 var globalTimer;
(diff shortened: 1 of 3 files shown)


js/interface.js
```diff
@@ -2189,9 +2182,13 @@ $allowBtnFilter.click(function(event) {
 
   if (value === 'tokens') {
     var tpl = Fliplet.Widget.Templates['templates.apiTokenList'];
+    var appTokens = _.groupBy(integrationTokenList, function (token) {
+      return _.get(_.first(token.apps), 'name', 'Other');
```
tonytlwu, line 2186:
```suggestion
      return _.get(_.first(token.apps), 'name', 'Other...');
```

js/interface.js
```diff
@@ -2189,9 +2182,13 @@ $allowBtnFilter.click(function(event) {
...
+    });
 
     $('.tokens-list').html(tpl({
-      integrationTokenList: integrationTokenList
+      // Sort by key (app name)
+      apps: _.fromPairs(_.sortBy(_.toPairs(appTokens), 1).reverse())
```
tonytlwu, line 2191: If we're using `_.sortBy()` but you've name one of the apps "Other", then we can't have "Other..." at the end. 

I also suggest sorting the string after they're all upper/lowercased so that A, B C don't come before a, b, c.

js/interface.js
```diff
@@ -2189,11 +2197,24 @@ $allowBtnFilter.click(function(event) {
...
     var tpl = Fliplet.Widget.Templates['templates.apiTokenList'];
+    var appTokens = _.groupBy(integrationTokenList, function (token) {
+      return _.get(_.first(token.apps), 'name', DESCRIPTION_APP_UNKNOWN);
+    });
+
+    // Sort by key (app name), but keep the unknown grouped tokens at the end of the list
+    var appsList = _.sortBy(_.mapValues(appTokens, function (tokens, name) {
+      return { name: name, tokens: tokens };
+    }), function (key) {
+      return key === DESCRIPTION_APP_UNKNOWN ? _.repeat('z', 32) : key;
```
tonytlwu, line 2208:
```suggestion
      return key === DESCRIPTION_APP_UNKNOWN ? 'z' : key.toUppercase();
```

...so that the sort is case-insensitive.

tonytlwu (changes requested):

PR #218, opened 2023-05-19
XOR-948 [FEAT] Added schema for data source

anerib:
### Product areas affected
 Data Source JSON Schema

### What does this PR do?

- Adds JSON schema for data source

### JIRA ticket
https://weboo.atlassian.net/browse/XOR-948

### Result
[screen recordings](https://weboo.atlassian.net/browse/XOR-948?focusedCommentId=45046)

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,162 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "title": "Data source component",
+  "description": "This is the schema for the data source object, including name, columns, access rules, and entries.",
+  "properties": {
+    "name": {
+      "title": "The name of the data source.",
+      "type": "string"
+    },
+    "columns": {
+      "title": "The list of columns included in the data source.",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "accessRules": {
+      "title": "The access rules for the data source.",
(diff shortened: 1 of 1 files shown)


schema.json
```diff
@@ -0,0 +1,162 @@
...
+          }
+        },
+        "required": ["type", "enabled"]
+      }
+    },
+    "entries": {
+      "title": "The entries (data) for the data source.",
+      "type": "array",
+      "items": {
+        "type": "object"
```
squallstar, line 157: - why does this say "object"?
- can you add a `data` Object attribute for each entry?

anerib: @squallstar , I tried adding it with `data` only as we discussed. However, it is being added as a `data` column in the data source. Please take a look at the screenshot below. 
image

As per the DS post API, the `entries` object should look like below:
```
"entries": [
    {
      "full name": "John Doe",
      "email": "johndoe@example.com"
    },
    {
      "full name": "Jane Doe",
      "email": "janedoe@example.com"
    }
  ]

squallstar: @anerib apologies you're right, I wasn't expecting this to go through a single API request — as I forgot we have an endpoint to create a data source including its initial data. What's the JS API you used for this?

squallstar (approved):

PR #226, opened 2023-12-27
ID-2816 [FEAT] pagination - tracking and saving column changes 

dwfliplet:


js/interface.js
```diff
@@ -473,6 +473,29 @@ function initializeTable(options) {
...
+    pageOffset,
+    sortConfig: sortConfig || getSortConfigForTable(getDataSourceQuery()),
+  });
+
+  updatePagination({
+    count: rows.length
+  });
+
+  const preparedData = table.prepareData(rows, columns);
+  HistoryStack.getCurrent().setData(preparedData);
```
tonytlwu, line 496: Why doesn't your linter add this automatically? There could be other errors in this PR that isn't easily detectable because the linter isn't working correctly.

Please investigate and fix your setup or the repository.

```suggestion

  HistoryStack.getCurrent().setData(preparedData);
```

js/history.js
```diff
@@ -3,17 +3,80 @@ Fliplet.Registry.set('history-stack', (function() {
...
   var currentIndex = 0;
 
+  const columnsInfo = {
+    original: () => stack[0].data[0],
+    reference: [],
+    getReference: () => {
+      if (!columnsInfo.reference.length) {
+        columnsInfo.reference = columnsInfo.original();
+      }
+      return columnsInfo.reference;
```
tonytlwu, line 13: See other comment re: linter setup

```suggestion

      return columnsInfo.reference;
```

js/history.js
```diff
@@ -126,6 +192,7 @@ Fliplet.Registry.set('history-stack', (function() {
     add: add,
     back: back,
     forward: forward,
-    getCurrent: getCurrent
+    getCurrent: getCurrent,
+    columnsInfo,
```
tonytlwu, line 196: It doesn't seem appropriate to be embedding a `columnsInfo` object into `HistoryStack` with its own methods when the majority of what's in `columnsInfo` doesn't seem to be interacting with the history stack.

I would suggest using a separate class `DSColumn` in a separate file to do this.

I do see that `original()` does use `stack[0].data[0]`, which references `stack` but that could be done through a new or existing public method in `HistoryStack`.

js/history.js
```diff
@@ -3,17 +3,80 @@ Fliplet.Registry.set('history-stack', (function() {
...
+    reference: [],
+    getReference: () => {
+      if (!columnsInfo.reference.length) {
+        columnsInfo.reference = columnsInfo.original();
+      }
+      return columnsInfo.reference;
+    },
+    current: [],
+    removedSinceReference: [],
+    removeColumns: (index, amount) => {
```
tonytlwu, line 17: I don't see this method being used

js/interface.js
```diff
@@ -901,10 +922,16 @@ function saveCurrentData() {
 
   var payload = getCommitPayload(entries);
 
+  const HistoryStack = Fliplet.Registry.get('history-stack');
+  const { deleteColumns, renameColumns } = HistoryStack.columnsInfo.getCommitPayload();
```
tonytlwu, line 926: I prefer making this part of the `getCommitPayload` function above so that it returns the `deleteColumns` and `renameColumns` properties as well

js/interface.js
```diff
@@ -606,9 +608,11 @@ function resetPagination() {
   pageSize = +$('#page-size').val() || undefined;
 }
 
-function fetchCurrentDataSourceEntries() {
-  resetPagination();
-
+function fetchCurrentDataSourceEntries(paginationReset = true) {
```
tonytlwu, line 611: 1. Use an `options` object so it's clearer in the caller context what the parameter is. `fetchCurrentDataSourceEntries(false)` isn't very obvious when reading what the `false` refers to.
2. `resetPagination` is a more suitable property name as it's an action (verb first) indicating it'll affect whether something is actioned or not

```suggestion
function fetchCurrentDataSourceEntries(options = {}) {
  if (options.paginationReset !== false)
```

js/interface.js
```diff
@@ -549,32 +563,20 @@ function updateDataSourceEntries() {
             });
           }
 
-          var columns = _.uniq(_.concat(currentDataSourceColumns, computedColumns));
+          const columns = columnsMismatched ? [...new Set([...currentDataSourceColumns, ...computedColumns])] : currentDataSourceColumns;
```
tonytlwu, line 566:
```suggestion
          const columns = columnsMismatched
            ? [...new Set([...currentDataSourceColumns, ...computedColumns])]
            : currentDataSourceColumns;
```

js/history.js
```diff
@@ -3,17 +3,80 @@ Fliplet.Registry.set('history-stack', (function() {
...
+  const columnsInfo = {
+    original: () => stack[0].data[0],
+    reference: [],
+    getReference: () => {
+      if (!columnsInfo.reference.length) {
+        columnsInfo.reference = columnsInfo.original();
+      }
+      return columnsInfo.reference;
+    },
+    current: [],
```
tonytlwu, line 15: `current` is only used internally within this class and not used publicly. So, I'd make it an private variable instead. e.g.

```
const current = [];

function moveColumns() {
  // This uses the current array
}

return {
  moveColumns
};
```

js/history.js
```diff
@@ -3,17 +3,80 @@ Fliplet.Registry.set('history-stack', (function() {
...
+      return collectionClone;
+    },
+    moveColumns: (columnOriginalIndexes, toIndex) => {
+      const reference = columnsInfo.getReference();
+      const current = columnsInfo.current;
+
+      columnsInfo.current = columnsInfo.moveColumnsInCollection(current, columnOriginalIndexes, toIndex);
+      columnsInfo.reference = columnsInfo.moveColumnsInCollection(reference, columnOriginalIndexes, toIndex);
+    },
+    getRenamedColumns: () => {
```
tonytlwu, line 42: This is another internal function. Check all returned properties/methods in the column class to make sure only the methods being used are returned -- unless they complement existing properties. e.g.

```
function addColumn() {
}

function removeColumn() {
}

return {
  addColumn, // Used outside of the class
  removeColumn // Not used outside of the class, but is a fair method to be available because it's complementary to addColumn
};
```

js/history.js
```diff
@@ -3,17 +3,80 @@ Fliplet.Registry.set('history-stack', (function() {
...
+
   function reset() {
     stack = [];
     currentIndex = 0;
   }
 
   // Clone data without losing the ID
   function cloneSpreadsheetData(data) {
-    return _.map(data, function(row) {
+    return _.map(data, function (row) {
```
tonytlwu, line 76: Something's wrong with your IDE setup for sure, and isn't following our linter rules. See https://docs.google.com/document/d/1XPlQHPbqp3rGGTekcC05bKtwKDDjrU2RMJ4VIHLXp64/edit#

```suggestion
    return _.map(data, function(row) {
```

js/history.js
```diff
@@ -41,6 +104,9 @@ Fliplet.Registry.set('history-stack', (function() {
     if (stack.length > 1) {
       currentIndex++;
     }
+    
+    // store columns info
```
tonytlwu, line 108: Start comments with correct capitalisation

```suggestion
    // Store columns info
```

js/interface.js
```diff
@@ -919,7 +946,9 @@ function saveCurrentData() {
     var clientIdMap = _.zipObject(clientIds, ids);
 
     cacheOriginalEntries(entries, clientIdMap);
-    table.setData({ columns: columns, rows: entries });
+    HistoryStack.columnsInfo.reset();
+
+    setTableData({ columns, rows: entries });
```
tonytlwu, line 951: This says that every time changes are saved, we are reinitialising the entire Handsontable instance, instead of using `hot.loadData()` to update existing data. Is this not a bit heavy-handed and potentially creates unnecessary initialisation time?

tonytlwu (changes requested):

js/interface.js
dwfliplet: @tonytlwu Yes, indeed, and it's done on purpose since Handsontable otherwise cannot handle the update of columns list with trailing column correctly. As we have pagination now, it's way cheaper to reinitialize Handsontable.

PR #230, opened 2024-04-18
fix(ID-4264): fixed updating DS order definition when column was removed or renamed

dwfliplet:
https://weboo.atlassian.net/browse/ID-4264


armine-fliplet (approved):
LGTM :+1:

simonchapman1986 (approved):
Risk: medium
Affected: datasources
Verify: order def when column removed/renamed

PR #269, opened 2026-05-21
feat: rename apps → projects (v3 user-visible wording)

hcarneiro:
## Summary

User-visible "Apps → Projects" rename to align with the v3 Studio rename shipped in Fliplet/fliplet-studio#8518.

## Scope (11 strings across 2 files)

- `interface.html`: "Show app's data sources", "Used by apps" (x2), "Apps will not be able to access", "All apps" / "Specific apps" rule scope, "Select apps affected by this rule"
- `js/interface.js`: "See all my app's data sources", "apps:" / "app:" inline references, "All apps" describer, "applied to all affected apps"

## What was kept

- Runtime end-user app references (e.g. "live users accessing the app", "app size", "make your app smaller") — these refer to the installed runtime app on a user's device
- CSS classes (`.apps-list`), data attributes (`data-app-source`, `data-apps`)
- API paths (`v1/apps/tokens`), Handlebars properties (`{{apps}}`, `{{#each apps}}`)
- `app-overlay` context flag (internal)

## Merge note

Includes a merge of `origin/production` with 1 conflict in `js/interface.js` (spreadsheet rebuild block) resolved by keeping the rename-branch HEAD which has the newer `page-loading-overlay` hide + `updatePaginationControls()` calls.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #270, opened 2026-05-21
feat: rename apps → projects (v3 user-visible wording)

hcarneiro:
## Summary

Replaces #269 with a clean branch based directly on `production`. Same rename work, but without the unrelated master-only commits (PS-1781 pagination, PS-864 security rules, Production #264) that were pulled in by the previous branch.

User-visible "Apps → Projects" rename to align with the v3 Studio rename shipped in Fliplet/fliplet-studio#8518.

## Scope (11 strings across 2 files)

- `interface.html`: "Show app's data sources", "Used by apps" (x2), "Apps will not be able to access", "All apps" / "Specific apps" rule scope, "Select apps affected by this rule"
- `js/interface.js`: "See all my app's data sources", "apps:" / "app:" inline references, "All apps" describer, "applied to all affected apps"

## What was kept

- Runtime end-user app references (e.g. "live users accessing the app", "app size", "make your app smaller") — these refer to the installed runtime app on a user's device
- CSS classes (`.apps-list`), data attributes (`data-app-source`, `data-apps`)
- API paths (`v1/apps/tokens`), Handlebars properties (`{{apps}}`, `{{#each apps}}`)
- `app-overlay` context flag (internal)

### DIFF
--- a/interface.html
+++ b/interface.html
@@ -14,7 +14,7 @@
             <div class="col-sm-8">
               <button data-create-source class="btn-round btn btn-primary">Create new data source</button>
               <button data-show-all-source class="btn-round btn btn-secondary hidden">Show all data sources</button>
-              <button data-app-source class="btn-round btn btn-secondary hidden">Show app's data sources</button>
+              <button data-app-source class="btn-round btn btn-secondary hidden">Show project's data sources</button>
             </div>
             <div class="flex col-sm-4">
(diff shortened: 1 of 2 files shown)


farhantariq12b (approved):
LGTM — clean rebase off production. `appPrefix` ternary updated for both branches, `data-apps` attribute selectors and API paths preserved.

PR #272, opened 2026-07-03
fix(PR-9): commit only changed rows from the Data Source Manager

zeryabkhan91:
## Summary
The Data Source Manager re-sent **every visible row** to `POST /commit` on every save — even with no changes — defeating the goal of PR-9. `getCommitPayload` compared the **whole entry** with `_.isEqual`, but the two sides are produced by different pipelines:

- `table.getData()` re-derives `order` as the **visual row index** and re-parses cell types (a stored number `30` comes back as the string `"30"`).
- The cached original held the **raw stored `order`** and raw types.

So an unchanged row never compared equal and was flagged as an update. This fixes it by comparing **normalized data content + position**, and caching the original `order` in the same visual-index coordinate `getData()` uses.

Fixes [PR-9](https://weboo.atlassian.net/browse/PR-9)

## Changes (all in `js/interface.js`)
- `cacheOriginalEntries` now caches `order` as the visual row index (matches `getData()`).
- New `normalizeValue` / `normalizeData` / `hasEntryChanged` helpers: scalar values coerced to strings, blanks dropped, keys sorted.
- `getCommitPayload` uses `hasEntryChanged()` (data content **or** position) instead of `_.isEqual(entry, original)`.

## Acceptance criteria
- [x] No-change save sends **0 rows** — `hasEntryChanged()` returns false when normalized data and order match
- [x] Number/string round-trip (`30` vs `"30"`) not flagged as a change
- [x] Blank cell `""` vs absent column not flagged (truthy-empty trap)
- [x] Reorder still persisted (moved rows detected via position)
- [x] Insert / update / delete behaviour unchanged

## Test plan
- [ ] Open a DS, **Save with no edits** → Network tab: `POST …/commit` body has `entries: []` (before: all rows sent)
- [ ] Edit one cell → exactly that row sent; drag-reorder rows → only moved rows sent
- [ ] A DS with numeric / boolean / blank cells → not falsely re-sent
- [ ] A DS whose stored `order` is non-sequential (imported / API-inserted) → no-change save sends `entries: []`

## Verification
- **No automated coverage on this branch** — `production` has no jest harness (`pagination.js` + tests were added by PS-1781, which is only on `master`). The pure diff logic was validated with a standalone script over 9 scenarios (no-change, type-coercion, boolean, empty/null, real change, cleared cell, reorder, key-order); all pass, and the old `_.isEqual` path was confirmed to flag the `30` vs `"30"` case. **Needs a manual QA pass** per the test plan.
- Lint: no new errors (`js/interface.js` has one pre-existing parse error on an unrelated arrow function, present on `production` before this change).

## Note on branches
This targets the `production` line (base `projects/PR-9`, cut from `production`), which does **not** contain PS-1781 pagination. The equivalent fix for the `master`/pagination line lives in `js/pagination.js` (PR #271). Both lines carry this bug in different forms and will need reconciliation when PS-1781 releases to production.

[PR-9]: https://weboo.atlassian.net/browse/PR-9?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -628,6 +628,28 @@ function removeEmptyColumnsInEntries(entries, emptyColumns) {
 }
 
 /**
+ * Serialise a value with object keys sorted at every depth, so two structurally
+ * equal objects/arrays always produce the same string regardless of key order.
+ * Plain JSON.stringify is key-order sensitive, which would spuriously re-commit a
+ * JSON/object cell whose keys came back from the grid in a different order.
+ * @param {*} value - Value to serialise
+ * @returns {String} Stable JSON serialisation
+ */
+function stableStringify(value) {
+  if (value === null || typeof value !== 'object') {
+    return JSON.stringify(value);
+  }
+
+  if (Array.isArray(value)) {
+    return '[' + value.map(stableStringify).join(',') + ']';
+  }
+
+  return '{' + Object.keys(value).sort().map(function(key) {
+    return JSON.stringify(key) + ':' + stableStringify(value[key]);
+  }).join(',') + '}';
+}
+
+/**
  * Normalise a single cell value for change-detection.
  * The spreadsheet round-trips every value through a text grid (getData +
  * parseCellValue), so a number stored as 30 comes back as the string "30",
@@ -642,9 +664,9 @@ function normalizeValue(value) {
     return '';
   }
 
-  // Arrays/objects: compare structurally via a stable serialisation
+  // Objects/arrays: compare structurally via a key-order-stable serialisation
   if (typeof value === 'object') {
-    return JSON.stringify(value);
+    return stableStringify(value);
   }
 
   return String(value);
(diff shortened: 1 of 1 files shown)


zeryabkhan91:
Addressed the two review findings in ee33263:

**Finding 1 (nested object/array key order)** — `normalizeValue` now serialises object/array cells via a new `stableStringify` that sorts object keys **at every depth** (arrays keep their order, since order is meaningful there). A JSON/object column whose nested keys come back from the grid in a different order is no longer spuriously re-committed — the residual false-positive is closed, not just documented.

**Finding 2 (dead `.sort()`)** — removed the top-level `Object.keys().sort()` in `normalizeData`; `hasEntryChanged` compares the maps with `_.isEqual`, which is already key-order-insensitive.

Re-validated the pure logic with the standalone script (now 10 scenarios incl. nested-key-order-irrelevant, nested-real-change, array-order-matters) — all pass; lint unchanged (only the pre-existing unrelated parse error). Per the review, QA should still exercise a **JSON/object column** alongside the non-sequential-`order` / numeric / blank / reorder rows.

PR #268, opened 2026-03-13
PS-1781 [Feat] Add pagination to data source manager

Komsomol:
## Summary
- Replaces unbounded `source.find({})` with paginated API queries (500 entries/page) to fix **502 errors and browser freezes** on data sources with 15K+ rows ([PS-1781](https://weboo.atlassian.net/browse/PS-1781))
- Uses existing backend pagination support — `POST /v1/data-sources/:id/data/query` with `limit`/`offset`, and `GET /v1/data-sources/:id?includeEntriesCount` for total count
- Adds Prev/Next pagination controls below the Handsontable grid with "1–500 of 15,234 entries" display
- Warns users about unsaved changes before navigating to another page
- Extracts pagination logic into a testable `js/pagination.js` module with 19 Jest tests

## Context
This is a fresh implementation of [ID-2816](https://weboo.atlassian.net/browse/ID-2816). The previous attempt (PR #217 by Tony, reverted in PR #227) was reverted after QA found 8+ bugs, many related to column management edge cases. This implementation deliberately avoids changing column management — it only paginates data loading and display.

## What changed
| File | Change |
|------|--------|
| `js/pagination.js` | New — pure functions for page computation and commit payload diffing |
| `js/interface.js` | Paginated fetch via `Fliplet.API.request`, pagination state, page nav handlers |
| `interface.html` | Pagination bar (Prev / info / Next) below spreadsheet |
| `css/interface.css` | Pagination bar styling |
| `widget.json` | Load `pagination.js` before `interface.js` |
| `test/pagination.test.js` | 19 tests — page info computation, commit payload edge cases |

## What's NOT changed (deliberately)
- **Save logic** — `getCommitPayload()` already only diffs against `entryMap.original`, which now contains only the current page. Saves are automatically page-scoped.
- **Column management** — untouched, works as before
- **Handsontable config** — unchanged, just receives fewer rows
- **Backend** — no API changes needed, all pagination endpoints already in production

## Test plan
- [ ] Open a data source with <500 entries — pagination bar should be hidden, behaviour unchanged
- [ ] Open a data source with 500+ entries — should see pagination bar with "1–500 of X entries"
- [ ] Click Next — loads next page, Previous enables
- [ ] Click Previous — returns to prior page
- [ ] Edit cells on a page, click Next without saving — should show "unsaved changes" confirmation
- [ ] Save changes on a page — should commit only that page's diff, refresh correctly
- [ ] Delete entries so last page becomes empty — page should clamp to the new last page
[...]

PR #274, opened 2026-07-16
ID-3689: Restore blank/empty data source manager render-race fix on master

farhantariq12b:
## JIRA Issue
https://weboo.atlassian.net/browse/ID-3689

## Why this exists
The original ID-3689 fix ([PR #249](https://github.com/Fliplet/fliplet-widget-data-source/pull/249)) was merged to `master` on 2025-09-10 and lived there ~2.5 months. It was then **silently dropped** when [PR #264](https://github.com/Fliplet/fliplet-widget-data-source/pull/264) merged `production` → `master` on 2025-11-25 (CodeRabbit's own summary of #264: *"Removed `waitUntilSized`/`renderSpreadsheet` helpers and switched requestAnimationFrame-based sequencing to `setTimeout(..., 0)`"*). The later pagination work ([PR #268](https://github.com/Fliplet/fliplet-widget-data-source/pull/268)) built on top of that already-regressed code, so `master` has shipped without the fix since.

This PR restores the fix on `master` **in lockstep with the production hotfix ([PR #273](https://github.com/Fliplet/fliplet-widget-data-source/pull/273))**, so the next release cut from `master` doesn't regress it a third time.

## Root cause (traced, not assumed)
- `.table-entries` has no CSS height — it's set imperatively by `windowResized()` (runs inside `Promise.all([entries, details]).then()` in `browseDataSource`).
- The entries table is built inside `fetchCurrentDataSourceEntries().then()` via `setTimeout(0)` — a separate, parallel promise chain.
- When entries resolves before details, the table is built **before** `windowResized()` sets the height → Handsontable measures a 0-height container → blank (`height:0` on `wtHolder`, as Tony diagnosed in 2023). Nondeterministic ordering = the ~50% "random" repro.

`waitUntilSized()` removes the ordering dependency entirely: it waits (via `requestAnimationFrame`) until `.table-entries` actually has dimensions before mounting the table.

## Changes
- Re-add `waitUntilSized()` and `renderSpreadsheet()` helpers.
- Both initial-load and subsequent-load branches now render via `renderSpreadsheet()`.
- **Adapted for pagination** (the difference vs. the production hotfix): `renderSpreadsheet()` keeps the `.page-loading-overlay` hide and `updatePaginationControls()` call inside the sized-gate, so paginated navigation is unaffected.

## Testing
- `node --check js/interface.js` — syntax valid
- `npm test` — **19/19 pagination tests pass** (unchanged; the fix doesn't touch `js/pagination.js`)
- Diff limited to the render path + two helper functions

## Testing notes
1. Open any app's Data Source Manager, open a data source → the entries table always renders with visible rows, never blank/empty.
2. Open a data source with multiple pages of entries, click Next/Prev and use the page-jump input → each page renders correctly, the "Loading page…" overlay clears, and pagination controls update.
3. With DevTools open and docked to the bottom, open a data source and resize the DevTools panel while it loads → table still renders instead of showing an empty/collapsed grid.
4. Close and reopen the DSM overlay a few times on the same data source → table renders every time.

## Risk Assessment
- Risk Level: LOW — restores an already-proven fix; single file; pagination behavior preserved and tested
- Files Modified: 1 (`js/interface.js`)
- Type: Bug fix (regression restoration)

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -269,6 +269,36 @@ function renderError(options) {
   });
 }
 
+function waitUntilSized(selector, callback) {
+  var el = document.querySelector(selector);
+
+  function check() {
+    if (!el || !document.contains(el)) {
+      return;
+    }
+
+    var rect = el.getBoundingClientRect();
+
+    if (rect.width > 0 && rect.height > 0) {
+      callback();
+    } else {
+      requestAnimationFrame(check);
+    }
+  }
+
+  check();
+}
+
+function renderSpreadsheet(rowsData) {
+  waitUntilSized('.table-entries', function() {
+    table = spreadsheet({ columns: columns, rows: rowsData });
+    $('.table-entries').css('visibility', 'visible');
+    $('.page-loading-overlay').addClass('hidden');
+    $('#versions').removeClass('hidden');
+    updatePaginationControls();
+  });
+}
+
 function fetchCurrentDataSourceDetails() {
   definitionEditor.setValue('');
   hooksEditor.setValue('');
@@ -538,28 +568,17 @@ function fetchCurrentDataSourceEntries(entries) {
 
       table = spreadsheet({ columns: columns, rows: [], initialLoad: true });
 
-      setTimeout(function() {
+      requestAnimationFrame(function() {
         table.destroy();
         initialLoad = false;
-
-        table = spreadsheet({ columns: columns, rows: rows });
-        $('.table-entries').css('visibility', 'visible');
-        $('.page-loading-overlay').addClass('hidden');
-
-        $('#versions').removeClass('hidden');
-        updatePaginationControls();
-      }, 0);
+        renderSpreadsheet(rows);
+      });
     } else {
       if (table) {
         table.destroy();
       }
 
-      table = spreadsheet({ columns: columns, rows: rows });
-      $('.table-entries').css('visibility', 'visible');
-      $('.page-loading-overlay').addClass('hidden');
-
-      $('#versions').removeClass('hidden');
-      updatePaginationControls();
+      renderSpreadsheet(rows);
     }
   })
     .catch(function onFetchError(error) {


js/interface.js
```diff
@@ -269,6 +269,36 @@ function renderError(options) {
...
+  var el = document.querySelector(selector);
+
+  function check() {
+    if (!el || !document.contains(el)) {
+      return;
+    }
+
+    var rect = el.getBoundingClientRect();
+
+    if (rect.width > 0 && rect.height > 0) {
```
zeryabkhan91: **Critical — this gate can never resolve on an ordinary error path, and dead-ends the DSM when it doesn't.**

The only thing that ever gives `.table-entries` a height is `windowResized()` (`js/interface.js:876`), which is called at `:907` inside `browseDataSource`'s `Promise.all([...]).then()`. `fetchCurrentDataSourceDetails()` (`js/interface.js:302`) has no `.catch` of its own, so when it rejects, `Promise.all` rejects, that `.then()` is skipped, and the terminal `.catch` at `:935-940` calls `getDataSources()` without sizing anything. The container stays at 0 height and this condition never becomes true, so `check()` re-arms via `requestAnimationFrame` forever and the callback never fires.

**Effect:** no spreadsheet is mounted at all and `#versions` stays hidden. During a page navigation it is worse — `goToPage()` (`:425-427`) disables prev/next/jump and shows the overlay, and `updatePaginationControls()` (`:406-408`) is the *only* thing that re-enables them, so both stay stuck. Recoverable only by resizing the window or reopening the DSM.

**Why it's reachable, not theoretical:** that `.catch`'s own comment names the trigger — *"EG: User try to edit an already deleted data source"* — and any network failure on `getById` does the same. Before this PR the same scenario rendered a blank-but-mounted table and cleared everything, so the user could carry on. #273 has a milder form (no table mounts, but no stuck overlay) because it doesn't move those two calls inside the gate.

**Fix — two small changes:**
1. add `windowResized();` to the `.catch` in `browseDataSource` (`js/interface.js:935`), so the container is always sized even when details fail;
2. give this gate a deadline and render anyway on expiry, hiding the overlay and re-enabling the controls on that path too.

Failing open to the old blank-table behaviour is strictly better than failing closed to a stuck overlay. Same fix is needed in #273.

js/interface.js
```diff
@@ -269,6 +269,36 @@ function renderError(options) {
...
+    } else {
+      requestAnimationFrame(check);
+    }
+  }
+
+  check();
+}
+
+function renderSpreadsheet(rowsData) {
+  waitUntilSized('.table-entries', function() {
```
zeryabkhan91: **Warning — no regression test on the changed path.**

The 19/19 that pass are pagination tests and don't touch this code, as the description says. jest is already configured (`package.json` `"test": "jest"`, `test/pagination.test.js`), and `waitUntilSized()` is close to ideal for a unit test: stub `document.querySelector`, `getBoundingClientRect` and `requestAnimationFrame`, then assert the callback fires once the rect is non-zero — and, per the `:282` comment, that it still fires on the timeout path.

This is the finding I'd most like addressed. The fix has now been silently lost twice through merges (#264 dropped it, #268 built on the regressed code); a test that fails on the pre-fix baseline is the only thing that reliably stops a third loss, and it's what makes this PR's stated purpose actually hold.

js/interface.js
```diff
@@ -269,6 +269,36 @@ function renderError(options) {
...
+    table = spreadsheet({ columns: columns, rows: rowsData });
+    $('.table-entries').css('visibility', 'visible');
```
zeryabkhan91: **Warning — half of #249 didn't come back, leaving this line as dead code.**

The original fix paired this with `$('.table-entries').css('visibility', 'hidden').attr('aria-busy', 'true');` at the top of the initial-load branch, and did `.css('visibility', 'visible').removeAttr('aria-busy')` here. Neither the hide nor the `aria-busy` was restored.

I grepped the branch to confirm the consequence: this is now the only `visibility` write in `js/interface.js`, and `css/interface.css` has no `.table-entries` visibility rule (the ones at 490/509 belong to `#toolbar .filter-form .find-controls`). So nothing ever sets it hidden — **this line is a no-op**, the empty `initialLoad: true` placeholder is visible during the gate where #249 hid it, and screen readers get no busy state.

For a PR whose stated job is re-adding the helpers, the dropped half should be a deliberate call — either restore it or note why it isn't wanted.

js/interface.js
```diff
@@ -538,28 +568,17 @@ function fetchCurrentDataSourceEntries(entries) {
 
       table = spreadsheet({ columns: columns, rows: [], initialLoad: true });
 
-      setTimeout(function() {
+      requestAnimationFrame(function() {
```
zeryabkhan91, line 559: **Warning — `setTimeout(0)` → `requestAnimationFrame` is an unlisted change.**

It matches #249 and #273 so it's consistent, but it isn't in the description's Changes section and it isn't needed: `waitUntilSized()` runs `check()` synchronously first and already rAFs when it has to wait, so this wrapper only adds a frame.

The behavioural delta is that rAF is suspended in background tabs, so opening the DSM in a hidden tab now leaves the destroyed placeholder in place until the tab is foregrounded. Self-healing, but a change. Either call `renderSpreadsheet(rows)` directly here or add a line to the description explaining the swap.

js/interface.js
```diff
@@ -538,28 +568,17 @@ function fetchCurrentDataSourceEntries(entries) {
...
-        $('.page-loading-overlay').addClass('hidden');
-
-        $('#versions').removeClass('hidden');
-        updatePaginationControls();
-      }, 0);
+        renderSpreadsheet(rows);
+      });
     } else {
       if (table) {
         table.destroy();
```
zeryabkhan91, line 567: **Warning — the destroy → recreate window is now unbounded.** *(possible; I haven't confirmed this by running it)*

`table.destroy()` runs synchronously here, but the reassignment happens inside the gate at `:294`. While it's pending, the module-level `table` (`:38`) points at a destroyed Handsontable, and it's dereferenced unguarded by `saveCurrentData()` (`:772` `table.onSave()`, `:775` `table.getData()`), `navigateToPage` (`:431` `table.hasChanges()`) and the close handlers at `:1337`/`:1865`. Pagination controls are disabled during navigation but the Save button isn't. Previously this gap was a single tick; now it's however long the gate takes.

Related: concurrent `renderSpreadsheet()` calls each start their own uncancelled rAF loop. The `fetchGeneration`/`thisFetch` guard at `:503` discards stale *fetches* but not stale *render gates*, so two waiters resolving in the same frame would both assign `table`, orphaning the first instance without `.destroy()`.

What would confirm it: open a data source with the container at 0 height and click Save while the gate is pending. Cheapest mitigation is a generation check inside the gate callback plus a null-guard on those `table` dereferences.

zeryabkhan91 (changes requested):
## Code review (via `/fliplet-pr-review`)

**Review basis:** `subject=4e96413d3a0811f263ca79dc86dd37030a6f3e27 standards=fliplet-engineering@0.15.5 revision=6e1ea86ad62667c1cc64926488a0146c05313143 source=fliplet-pr-review digest=sha256:b54eff56f86bcdda`

### Summary

Restores the ID-3689 fix on `master` after #264 dropped it. **The diagnosis is right and I verified it independently** rather than taking the body's word for it: `.table-entries` has no CSS height (`css/interface.css:229-233`), it is sized imperatively by `windowResized()` (`js/interface.js:876`), and it is called from `browseDataSource`'s `Promise.all(...).then()` at `:903-907` while the table is built on a separate chain — which is the same mechanism Tony reached on ID-3689 in Nov 2023 (`wtHolder` with `height: 0`). I also confirmed the regression itself: `waitUntilSized` is absent from **both** `master` and `production` right now, so the fix is live on neither branch.

I diffed this against the production sibling #273 too — #274 is #273 plus exactly the two pagination lines inside `renderSpreadsheet`, as the description claims. No drift between them.

### Regression risk

Low-to-moderate, Studio-only, no runtime/app-facing or auth surface. The collateral worth watching is that this PR moves the `.page-loading-overlay` hide and `updatePaginationControls()` **inside** an unbounded wait gate, so any path where the gate never resolves turns today's "blank table you can work around" into a dead-ended UI. That trade is specific to #274 — #273 keeps neither call inside the gate.

### The blocking issue, in one place

The inline comment on `js/interface.js:282` has the detail, but the key part involves a line outside this diff, so repeating it here:

`fetchCurrentDataSourceDetails()` (`js/interface.js:302`) has no `.catch` of its own. When it rejects, `Promise.all` rejects, the `.then()` holding `windowResized()` is skipped, and the terminal `.catch` at `:935-940` calls `getDataSources()` **without** sizing anything. `.table-entries` therefore never gets a height and the new gate polls forever. The trigger is ordinary — that `.catch`'s own comment names it, *"EG: User try to edit an already deleted data source"* — plus any network failure on `getById`.

Smallest fix is two changes: add `windowResized();` to that `.catch`, and give `waitUntilSized()` a deadline so it renders anyway after ~1-2s instead of polling forever. Failing open to the old blank-table behaviour beats failing closed to a stuck overlay.

### Acceptance criteria (ID-3689)

The ticket is a 2023 bug report with no formal AC list. Against the reported behaviour: the blank-render fix ✅ (`:288-297` removes the ordering dependency), Illia's ~50% DevTools-resize repro ✅ (the gate re-checks every frame), and "survives on the branch that ships" ⚠️ — satisfied only once #273 lands, since this widget deploys from `production` and #274 alone changes nothing for users.

Worth noting on the ticket side: ID-3689 currently reads **Done (Ready for release)** while the code is on neither branch, and Ian has asked for status four times (09 Jul, 16 Jul, 03 Aug, and again today).

### CI status

**Green** — Codacy Static Code Analysis pass, the only configured check.

### Verdict — REQUEST_CHANGES

To be clear about proportion: this is close to approvable and I would rather it landed this week than sat another two months. The change is small, correctly targeted at `master` for a widget repo, the ES5 down-conversion from #249's `const`/arrow syntax is the right call for this file, and Codacy is clean.

What holds it is one confirmed error path that dead-ends the DSM (inline on `:282`), which is genuinely small to fix. The other four are warnings, not blockers — though the missing test is the one I would most like to see, given this fix has now been silently reverted twice.

Both fixes belong in #273 as well, which is the half that actually reaches users.

farhantariq12b:
Pushed 48639ff addressing all points from @zeryabkhan91's review:

- **Blocking (stuck gate on error path):** `browseDataSource`'s terminal `.catch` now calls `windowResized()` so `.table-entries` always gets sized even when `fetchCurrentDataSourceDetails()` rejects, and `waitUntilSized()` now takes a 2s timeout and falls open (renders anyway) instead of polling `requestAnimationFrame` forever.
- **Missing test:** extracted `waitUntilSized()` into `js/waitUntilSized.js` (same CommonJS-export-for-testing pattern as `js/pagination.js`) with `test/waitUntilSized.test.js` covering immediate resolution, polling-until-sized, DOM-removal short-circuit, and the new timeout fallback. 23/23 tests pass.
- **Dropped visibility/aria-busy pairing:** restored `$('.table-entries').css('visibility', 'hidden').attr('aria-busy', 'true')` on the initial-load branch, so the `visibility:visible` / `removeAttr('aria-busy')` in `renderSpreadsheet` is no longer a no-op.
- **Unbounded destroy→recreate window:** added a `fetchGeneration` check inside the render gate's callback so a stale render (superseded by a newer fetch while the gate was waiting for sizing) can't assign `table` out from under a newer instance.
- **Unlisted `setTimeout(0)` → `requestAnimationFrame` swap:** left as-is per your note that it matches #249/#273 — calling that out here for the record since it wasn't in the original description.

Same fixes need to land in #273 next since that's the branch that actually ships.

farhantariq12b:
Pushed 0b9aa84 covering the remaining warning: null-guarded the `table` dereferences during the destroy→recreate window (`navigateToPage`, `saveCurrentData()`, the `[data-save]` handler, and `shown.bs.tab`) and set `table = null` right after both `table.destroy()` calls that precede a `renderSpreadsheet()` call, so the gap is detectable. All treat a null `table` as "nothing to save / no changes" rather than throwing. 23/23 tests still pass.

zeryabkhan91 (approved):
LGTM

PR #276, opened 2026-08-20
PS-1781 [Fix] Only commit row order when the user reordered rows

zeryabkhan91:
## Why

[PS-1781](https://weboo.atlassian.net/browse/PS-1781) — London Doctors Clinic get a **502 Bad Gateway** when adding or removing rows in their ~15,000-row "Registration" data source. Reported March, re-reported 19 Aug, now High.

**The 502 is on the commit, not the fetch.** That matters, because the assumption so far has been that the payload problem is on load.

`getData()` re-derives each entry's `order` from its **visual row index**, and `getCommitPayload()` compares the whole entry against the cached original with `_.isEqual`. Deleting one row shifts the index of every row beneath it, so all of them compare unequal and get re-sent as updates.

Measured against a real 15,000-row data source on a local stack:

| user action | entries sent | body | time |
|---|---|---|---|
| delete one row near the top | 14,994 | 3.22 MB | **14.6 s** |
| the same delete, order excluded | 0 | 143 B | **0.19 s** |

14.6 s against an idle local database is comfortably past a production gateway timeout, which is the 502. It also explains the detail in the ticket that everyone skipped: *"deleting 1 user at a time works with no errors"* — deleting the **last** row shifts nothing, so nothing is re-sent.

## What changes

Row order is only compared when the user actually **dragged a row during that save**, tracked by a `rowsMoved` flag set in `afterRowMove` and cleared in `reset()`.

The cached original's `order` is also changed from the **stored value** to the row's **position in the loaded list**, which is the coordinate `getData()` re-derives. Without that, any data source whose stored order has gaps or nulls — imports, API inserts, anything previously deleted from — compares unequal on every row regardless.

Same 15,000-row data source, entries sent in the commit:

| action | before | after |
|---|---|---|
| save with no edits | 15,000 | **0** |
| delete one row | 14,994 | **0** |
| delete 10 scattered rows | 14,990 | **0** |
| edit one cell | 15,000 | **1** |
| insert one row | 15,000 | **1** |
| drag row 10 → 20 | 15,000 | **11** |

A reorder still commits exactly the rows whose position changed, so drag-to-reorder keeps working — it is not disabled to buy the speed.

## Scope

Deliberately narrow, and it is **not** the pagination work.

It does not change how entries are loaded, so a 15,000-row data source still fetches every row when opened. That is the other half of PS-1781 — the multi-MB payload and DOM freeze from the original analysis — and is handled by pagination in **#275**. The two are complementary: this fixes why the *save* is huge, pagination fixes why the *load* is huge. This one is `O(rows the user changed)` rather than `O(page size)`, so it is the better fix for the 502 specifically, and it is small enough to ship on its own.

## Relationship to #272

[#272](https://github.com/Fliplet/fliplet-widget-data-source/pull/272) (PR-9) targets the same code and is worth landing, but it does **not** fix this. Its goal is that a *no-change* save sends 0 rows. I ran its logic against this scenario: because it still compares position unconditionally, `delete row #6` sends **14,994** entries and `delete row #7500` sends **7,499**. It removes the idle waste, not the shift cascade.

If both land, `hasEntryChanged` should end up with #272's value normalisation (type round-trips like `30` vs `"30"`, blanks, stable JSON key order) **and** this PR's conditional position check. They compose cleanly; this PR keeps the comparison deliberately simple so the merge is easy.

## Testing

No automated coverage on this line — `production` has no jest harness (`pagination.js` and its tests came with PS-1781 and are only on the pagination branch). The comparison logic was validated with a standalone script across the eight scenarios in the table above, and the 3.22 MB / 14.6 s vs 143 B / 0.19 s figures come from real `POST /commit` calls against a seeded 15,000-row data source.

Lint: identical problem count to `production` (13 pre-existing, none added).

Manual QA:

1. **The bug** — open a data source with 15,000+ rows, delete a handful of rows from the middle, save. Expect a fast save and no 502. Check the Network tab: the commit body should contain `entries: []` and only the deleted ids.
2. **Reorder still persists** — drag a row to a new position, save, reload. Expect it to stay where you put it, and only the moved rows in the commit body.
3. **Edits** — change one cell, save. Exactly one entry in the body.
4. **Inserts** — add a row at the end, save, reload. It should persist and stay last.
5. **No-change save** — open and save without editing. `entries: []`.
6. **A data source with gaps** — one that has had rows deleted before, or was imported. A no-change save should still send `entries: []`.
7. **Small data source** — unchanged behaviour throughout.

[PS-1781]: https://weboo.atlassian.net/browse/PS-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/interface.js
+++ b/js/interface.js
@@ -314,12 +314,20 @@ function fetchCurrentDataSourceUsers() {
 function cacheOriginalEntries(entries, clientIdMap) {
   entryMap.original = {};
 
-  _.forEach(entries, function(entry) {
+  _.forEach(entries, function(entry, index) {
     if (!entry.id && typeof clientIdMap === 'object') {
       entry.id = clientIdMap[entry.clientId];
     }
 
-    entryMap.original[entry.id] = _.pick(entry, ['id', 'data', 'order']);
+    // `order` is cached as the row's position in this list, not its stored value,
+    // because that is the coordinate getData() re-derives at save time. Caching the
+    // stored value makes every row of an imported or gap-ridden data source compare
+    // unequal.
+    entryMap.original[entry.id] = {
+      id: entry.id,
+      data: entry.data,
+      order: index
+    };
   });
 }
 
@@ -623,8 +631,38 @@ function removeEmptyColumnsInEntries(entries, emptyColumns) {
  * @param {Array} entries - Latest entries to be committed
  * @returns {Object} List of new/updated entries and deleted IDs
  */
+/**
+ * Determine whether an entry needs committing.
+ *
+ * Position is only compared when the user actually dragged a row. getData()
+ * re-derives `order` from the visual row index, so deleting one row shifts the
+ * index of every row beneath it and a plain comparison flags them all as changed
+ * - on a 15,000-row data source that is a ~3 MB commit that takes ~15s and times
+ * out as a 502 (PS-1781). Skipping the position check unless rows were moved
+ * keeps a delete to just the delete, while a genuine reorder still commits the
+ * rows whose position changed.
+ * @param {Object} entry - Current entry from the table
+ * @param {Object} original - Cached original entry
+ * @param {Boolean} comparePosition - Whether row order is meaningful for this save
+ * @returns {Boolean} True when the entry should be committed
+ */
+function hasEntryChanged(entry, original, comparePosition) {
+  if (!_.isEqual(entry.data, original.data)) {
+    return true;
+  }
+
+  if (!comparePosition) {
+    return false;
+  }
+
+  return typeof entry.order !== 'undefined' && entry.order !== original.order;
+}
+
 function getCommitPayload(entries) {
   entries = entries || [];
+
+  // Row order only matters to this save if the user dragged a row during it
+  var comparePosition = !!(table && typeof table.hasRowsMoved === 'function' && table.hasRowsMoved());
 
   var inserted = [];
   var updated = [];
@@ -663,7 +701,7 @@ function getCommitPayload(entries) {
       return;
     }
 
-    if (_.isEqual(entry, original)) {
+    if (!hasEntryChanged(entry, original, comparePosition)) {
       return;
     }
 
--- a/js/interface.js
+++ b/js/interface.js
@@ -314,19 +314,17 @@ function fetchCurrentDataSourceUsers() {
 function cacheOriginalEntries(entries, clientIdMap) {
   entryMap.original = {};
 
-  _.forEach(entries, function(entry, index) {
+  _.forEach(entries, function(entry) {
     if (!entry.id && typeof clientIdMap === 'object') {
       entry.id = clientIdMap[entry.clientId];
     }
 
-    // `order` is cached as the row's position in this list, not its stored value,
-    // because that is the coordinate getData() re-derives at save time. Caching the
-    // stored value makes every row of an imported or gap-ridden data source compare
-    // unequal.
+    // `order` is the value the server stores, not a visual index. It is only used
+    // when the user reorders, to work out which rows genuinely need renumbering.
     entryMap.original[entry.id] = {
       id: entry.id,
       data: entry.data,
-      order: index
+      order: entry.order
     };
   });
 }
@@ -634,35 +632,38 @@ function removeEmptyColumnsInEntries(entries, emptyColumns) {
 /**
  * Determine whether an entry needs committing.
  *
- * Position is only compared when the user actually dragged a row. getData()
- * re-derives `order` from the visual row index, so deleting one row shifts the
- * index of every row beneath it and a plain comparison flags them all as changed
- * - on a 15,000-row data source that is a ~3 MB commit that takes ~15s and times
- * out as a 502 (PS-1781). Skipping the position check unless rows were moved
- * keeps a delete to just the delete, while a genuine reorder still commits the
- * rows whose position changed.
+ * Position is never compared. getData() derives `order` from the visual row index,
+ * so deleting one row shifts the index of every row beneath it and a comparison
+ * that included position flagged them all as changed - on a 15,000-row data source
+ * that is a ~3 MB commit taking ~15s, which times out as a 502 (PS-1781).
+ *
+ * When the user does drag a row, position is compared against the row's *stored*
+ * order, so only the rows that genuinely need renumbering are committed.
  * @param {Object} entry - Current entry from the table
  * @param {Object} original - Cached original entry
- * @param {Boolean} comparePosition - Whether row order is meaningful for this save
+ * @param {Boolean} rowsMoved - Whether the user dragged a row during this save
  * @returns {Boolean} True when the entry should be committed
  */
-function hasEntryChanged(entry, original, comparePosition) {
+function hasEntryChanged(entry, original, rowsMoved) {
   if (!_.isEqual(entry.data, original.data)) {
     return true;
   }
 
-  if (!comparePosition) {
+  if (!rowsMoved) {
     return false;
   }
 
+  // A reorder is the one case where position must be written. Comparing the new
+  // dense index against the stored value keeps the commit to the rows that
+  // actually need renumbering rather than the whole data source.
   return typeof entry.order !== 'undefined' && entry.order !== original.order;
 }
 
 function getCommitPayload(entries) {
   entries = entries || [];
 
-  // Row order only matters to this save if the user dragged a row during it
-  var comparePosition = !!(table && typeof table.hasRowsMoved === 'function' && table.hasRowsMoved());
+  // Position is only worth writing when the user dragged a row during this save
+  var rowsMoved = !!(table && typeof table.hasRowsMoved === 'function' && table.hasRowsMoved());
 
   var inserted = [];
   var updated = [];
@@ -701,7 +702,7 @@ function getCommitPayload(entries) {
       return;
     }
 
-    if (!hasEntryChanged(entry, original, comparePosition)) {
+    if (!hasEntryChanged(entry, original, rowsMoved)) {
       return;
     }
 
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -886,7 +886,14 @@ function spreadsheet(options) {
             entry.data[header] = visualRow[index];
           }
 
-          entry.order = order;
+          // Only stamp a position when the user actually reordered rows. This is a
+          // dense 0-based index, which is incompatible with stored orders that are
+          // sparse or null, so writing it on a data-only edit or an insert would
+          // silently move the row. A reorder renumbers the whole list, so there the
+          // dense value is correct for every row.
+          if (rowsMoved) {
+            entry.order = order;
+          }
 
           // Only parse the column value when required
           if (options.parseJSON && typeof entry.data[header] === 'string') {
--- a/js/spreadsheet.js
+++ b/js/spreadsheet.js
@@ -14,6 +14,9 @@ function spreadsheet(options) {
   var columnNameCounter = 1; // Counter to anonymous columns names
   var rendered = 0;
   var isDestroyed = false; // Flag to prevent actions after instance is destroyed
+  // Set once the user drags a row. Row order is only worth committing when they
+  // actually reordered something - see getData().
+  var rowsMoved = false;
 
   /**
    * Given an array of data source entries it does return an array
@@ -578,6 +581,7 @@ function spreadsheet(options) {
       }
     },
     afterRowMove: function() {
+      rowsMoved = true;
       onChange();
     },
     afterCreateRow: function() {
@@ -987,6 +991,7 @@ function spreadsheet(options) {
   function reset(resetHistory) {
     search('clear');
     setChanges(false);
+    rowsMoved = false;
 
     $('.save-btn').addClass('hidden');
     $('.data-save-status').addClass('hidden');
@@ -1027,6 +1032,9 @@ function spreadsheet(options) {
     onSaveError: onSaveError,
     hasChanges: hasChanges,
     setChanges: setChanges,
+    hasRowsMoved: function() {
+      return rowsMoved;
+    },
     onChange: onChange
   };
 }


Arpanexe:
Review findings:

- **[P1] Data-only edits can unexpectedly reorder sparse data sources.** `getData()` assigns every entry a dense visual `order`. When only `data` changes, the full entry—including that recalculated order—is still added to the update payload at [`js/interface.js:708`](https://github.com/Fliplet/fliplet-widget-data-source/blob/f4d0a131c2730ebf7d4cd5cbe78e4f6e7a123a89/js/interface.js#L708). For stored orders `[0, 10, 20]`, editing the last row sends `order: 2`; after reload it sorts before the row at `10`. Rows with `null` order can jump similarly. Omit `order` from data-only updates or retain the stored order unless reordering.

- **[P1] Inserted rows use an incompatible order coordinate.** New entries are sent with their dense visual index at [`js/interface.js:678`](https://github.com/Fliplet/fliplet-widget-data-source/blob/f4d0a131c2730ebf7d4cd5cbe78e4f6e7a123a89/js/interface.js#L678), while existing sparse/null orders remain unchanged. Appending to `[0, 10, 20]` sends `order: 3`, so the new row reloads second instead of last. This needs insertion tests covering gapped and null orders.

- **[P2] Combining deletion with a row drag recreates the oversized commit.** The save-wide boolean at [`js/interface.js:665`](https://github.com/Fliplet/fliplet-widget-data-source/blob/f4d0a131c2730ebf7d4cd5cbe78e4f6e7a123a89/js/interface.js#L665) enables position comparison for every row. Delete near the top, then drag any row before saving, and all rows shifted by the deletion become updates again—potentially restoring the 15,000-row payload and 502. Tracking the actually moved range/IDs would avoid conflating deletion shifts with reordering.

No CI checks or automated tests were reported for the head commit.

zeryabkhan91:
Thanks Arpan — all three reproduce. Fixed in `f5a8d25`.

You were right that the root problem was broader than the comparison: the fix stopped *comparing* `order` but the payload still *carried* it, and a dense visual index is simply not compatible with a sparse or null stored order.

**P1 #1 — data-only edits reordering sparse data sources.** Reproduced against the API on a source stored as `[0, 10, 20]`. Editing row C sent `order: 2` and the read came back `A, C edited, B` — exactly as you described.

`getData()` now only stamps `order` when the user actually dragged a row, so a data-only edit carries no `order` field at all. Same edit after the fix:

```
before:  0 | A     10 | B     20 | C
after:   0 | A     10 | B     20 | C edited     ← stored order survives
```

**P1 #2 — inserts using an incompatible coordinate.** Also reproduced: appending to `[0, 10, 20]` with `order: 3` reloaded second. New rows now carry no `order`, so the column stays `null` and sorts last:

```
0 | A     10 | B     20 | C edited     null | NEW     ← correct append
```

**P2 #3 — delete plus drag re-enabling position comparison for every row.** Replaced the save-wide boolean. Position is now compared against the row's **stored** order rather than a flag that applies to the whole save, so a pure drag of row 10 → 20 on a 15,000-row data source commits **11 rows instead of 15,000**.

Rows sent on a 15,000-row data source:

| action | before | after |
|---|---|---|
| save, no edits | 15,000 | 0 |
| delete one row | 14,994 | 0 |
| delete 10 scattered rows | 14,990 | 0 |
| edit one cell | 15,000 | 1 |
| insert one row | 15,000 | 1 |
| drag row 10 → 20 | 15,000 | 11 |

**One limitation I have not solved, flagging rather than hiding it:** deleting a row *and* dragging another **in the same save** still commits every row below the delete — 14,994 in that scenario. Their indices genuinely did all shift, and a dense sequence has to be rewritten to stay consistent. It is no worse than current production, where *every* save commits every row, but it is not fixed. The proper answer is to slot a moved row between its neighbours' stored orders instead of renumbering densely, which removes the whole class of problem — worth its own ticket rather than growing this one.

**On the missing CI:** correct, and it is not something this branch can fix. `production` has no jest harness — `pagination.js` and its tests arrived with PS-1781 and exist only on the pagination line. The logic here was validated with a standalone script across the scenarios in the table, plus the live API checks quoted above against a seeded sparse data source. Adding jest to the `production` line would be a reasonable separate PR if you want automated coverage before this merges.

Lint is unchanged from `production` (13 pre-existing problems, none added). There is a QA test plan for this PR covering the delete cases, reorder persistence, and sparse/gapped sources — happy to attach it here or to the ticket.

zeryabkhan91:
Delete + drag is fixed too — `b7cc32e`.

The cause was renumbering from zero. A dense index comes from the visual position, so a delete shifts every row beneath it and they all differ from their stored value even though nothing about them moved.

Reordering now **redistributes the order values the rows already hold** across the current visual sequence, instead of assigning `0..n`. That changes the arithmetic:

- **Deleting needs no rewrite at all.** The surviving values are still ascending, so every row is handed back the value it already had.
- **A drag only rewrites the span it crossed**, even when a delete in the same save shifted every index below it.

| scenario | before | now |
|---|---|---|
| delete one row, no drag | 0 | 0 |
| **delete one row + drag, same save** | **14,994** | **12** |
| pure drag of row 10 → 20 | 11 | 11 |
| reorder flag set but nothing moved | 15,000 | 0 |

Sparse orders survive, because the pool is whatever the data source already used. Dragging the first row of a `0/10/20` source to the end writes `B=0, C=10, A=20` — gaps preserved, verified against the API:

```
start:  0 | A    10 | B    20 | C
drag A to the end
  ->    0 | B    10 | C    20 | A
```

If any row has no stored order the pool cannot be trusted, so a dense sequence is assigned instead — which also heals a data source that was never ordered.

**Still unavoidable:** moving the first row to the very end rewrites everything, because every position genuinely does change. That is arithmetic rather than a design flaw, and it is bounded by the size of the move.

Lint is back to parity with `production` (13 problems, none added) — dropping the order stamp left an unused `forEach` index argument, now removed.

zeryabkhan91:
Automated tests added — `d4bd667`. That closes the last point from your review.

There was nowhere to put them before: the comparison lived inside `interface.js`, which depends on the DOM, lodash and Fliplet globals, and this line has no jest harness (`pagination.js` and its specs arrived with PS-1781 and exist only on the pagination branch).

The change detection is now **`js/entry-diff.js`** — pure functions with the comparison helpers injected rather than imported, mirroring how `pagination.js` is structured on the other line. `getCommitPayload()` is a thin wrapper over it, so behaviour is unchanged.

**25 specs**, covering everything this PR has been through:

| area | covered |
|---|---|
| the original 502 | no-change save, delete at the top, deletes from the middle, single and multiple cell edits |
| **P1 #1** | data-only edits carry no `order`, on sparse and dense sources alike |
| **P1 #2** | inserts carry no `order` so they sort last; recovered ids; insert alongside a delete |
| **P2 #3** | a drag rewrites only the span it crossed; delete + drag stays bounded; a flag with no actual move writes nothing; sparse pools reused; resulting sequence strictly ascending |
| edge cases | null entry list, emptying the table, missing stored order falling back to a dense sequence |

**I checked the suite actually catches the bugs** rather than just passing. Run against a reconstruction of the pre-fix behaviour — dense index always stamped, position always compared — **12 assertions fail**, including:

```
sends only the delete when a row near the top is removed
   length 14994 != 0
stays bounded when a delete and a drag happen in the same save
   14994 not < 20
does not attach an order when only the data changed
   Expected values to be strictly equal
```

It fails on the bug and passes on the fix.

**Two things you should know before merging:**

1. **`package-lock.json` is not regenerated.** `npm install` cannot reach the registry from this machine (three attempts, socket timeouts), so the `jest` devDependency is declared but the lock still needs refreshing by someone who can install. If anything runs `npm ci` it will complain until then.
2. **jest itself has not been run here** — the specs were executed under an equivalent `assert` harness. The first real `jest` run will be whoever installs first.

Lint is unchanged from `production` (13 pre-existing problems), and both new files are clean.

Arpanexe:
Re-review findings against current head `7ab3bab`:

- **[P1] Inserts no longer preserve their visual position.** New rows are always committed without `order` at [`js/entry-diff.js:181`](https://github.com/Fliplet/fliplet-widget-data-source/blob/7ab3babed37774333480fc4beaa15f944b20f5b1/js/entry-diff.js#L181), despite the UI supporting insert-before/after. A middle insertion reloads at the end; on an all-null source, an appended row sorts first because the API orders equal/null values by descending ID. Multiple inserts can also reverse. The tests only verify that `order` is absent, not the post-API row sequence.

- **[P1] Reordering a null-order source still recreates the 15,000-row payload.** If any stored order is null, [`js/entry-diff.js:108`](https://github.com/Fliplet/fliplet-widget-data-source/blob/7ab3babed37774333480fc4beaa15f944b20f5b1/js/entry-diff.js#L108) replaces every order with a dense sequence. Running the current module against 15,000 null-order rows produced **15,000 updates for one drag**, risking the same timeout the PR addresses.

- **[P1] `package-lock.json` is not synchronized.** Jest was added at [`package.json:23`](https://github.com/Fliplet/fliplet-widget-data-source/blob/7ab3babed37774333480fc4beaa15f944b20f5b1/package.json#L23), but the tracked lockfile contains no Jest dependency. `npm ci` will fail, and the Jest suite has not actually been executed.

The original data-only edit issue is fixed, and delete-plus-drag is bounded for numeric order pools.

There are no GitHub Actions runs. The external [Codacy check](https://github.com/Fliplet/fliplet-widget-data-source/runs/96706220218) is currently `action_required`, with no annotations exposed.

zeryabkhan91:
All three fixed — `3fcfd27`. Each one reproduced first.

**P1 — inserts lost their position.** Confirmed, and worse than the report: the API default sort is `[order ASC, id DESC]`, so among equal or null orders it reads **newest first**. Inserting A, B, C into an unordered data source read back as **C, B, A**.

Two changes. The widget now asks for `[order ASC, id ASC]` explicitly rather than relying on that default. And a new row is slotted into the gap between its neighbours' stored orders; where they leave no room the following row is renumbered too, so the write stays bounded and rows *above* the insertion point are never touched. A row is left unnumbered only when nothing on either side is numbered — any number there would sort it ahead of every existing row, which is the bug you spotted on the all-null source.

| case | result | writes |
|---|---|---|
| sparse `0/10/20`, insert after A | `A,NEW,B,C` | 1 |
| dense `0/1/2`, insert after A (no gap) | `A,NEW,B,C` | 2 |
| insert at the very top | `NEW,A,B,C` | 1 |
| append ×2 | `A,B,C,N1,N2` | 2 |
| unordered source, append ×2 | `A,B,C,N1,N2` | 2 |

**P1 — reordering an unordered data source.** Confirmed at 15,000. Those rows read back on their ids, so only the prefix up to the last row that actually moved needs numbering; the rest still read correctly untouched.

| | before | after |
|---|---|---|
| unordered 15k, drag 10 → 20 | 15,000 | **21** |
| unordered 15k, flag set but nothing moved | 15,000 | **0** |
| numeric 15k, drag 10 → 20 | 11 | 11 |
| numeric 15k, delete + drag | 12 | 12 |
| numeric 15k, delete only | 0 | 0 |

Moving the first row to the very end still rewrites everything — unavoidable, every position genuinely changes.

**P1 — `package-lock.json`.** You were right, and I could not generate a trustworthy lockfile here (no registry access). So rather than commit one I cannot verify, **jest is dropped entirely** and the specs run on Node's built-in runner:

```
"test": "node --test test/*.test.js"
```

No test dependencies at all, `package.json` and the lockfile agree again, `npm ci` works — and the suite genuinely runs now rather than being asserted under a stand-in:

```
# tests 42
# pass 42
# fail 0
```

**On the test quality point** — fair, and fixed. The insert specs now apply the payload and read the rows back in sort order, asserting the resulting sequence rather than that a field is absent. I checked they actually bite: reverting the insert placement fails 3 of them.

Lint is back at parity with `production` (13 pre-existing, none added) — the first cut of this added a `no-loop-func` error, now gone.

**Two things I have not solved**, flagged rather than hidden:

1. A **middle** insert into a data source with no stored order still lands at the end. Placing it properly means numbering every row, which is the payload size this PR exists to avoid.
2. Mixed data sources — some rows numbered, some null — are not handled precisely by the insert placement. Rare in practice (a data source is normally all one or the other), but it is a real gap.

Arpanexe:
Re-review findings against current head `3fcfd27`:

- **[P1] Unordered inserts are only fixed inside the Data Source Manager.** Rows left unnumbered at [`js/entry-diff.js:216`](https://github.com/Fliplet/fliplet-widget-data-source/blob/3fcfd27f4a137b3c858261cb320de627a11ae5e0/js/entry-diff.js#L216) appear last only because the manager now requests `id ASC` at [`js/interface.js:383`](https://github.com/Fliplet/fliplet-widget-data-source/blob/3fcfd27f4a137b3c858261cb320de627a11ae5e0/js/interface.js#L383). Default API/application queries still use `id DESC`, so an appended null-order row appears first to consumers while appearing last in the manager. Its intended position was never persisted. Middle inserts into an all-null source also still reload at the end.

- **[P1] Packed insertion creates duplicate orders and relies on IDs to repair them.** The widening condition at [`js/entry-diff.js:210`](https://github.com/Fliplet/fliplet-widget-data-source/blob/3fcfd27f4a137b3c858261cb320de627a11ae5e0/js/entry-diff.js#L210) considers only the insert count, not the displaced rows. Inserting into orders `0/1/2` produces `NEW=1`, displaced `B=2`, while `C` remains `2`. With `B.id=300` and `C.id=200`, the manager reloads `A,NEW,C,B`. Default consumers using `id DESC` can produce a different sequence again. Assigned orders need to remain unique or independent of ID tie-breaking.

- **[P1] Mixed numeric/null sources can still recreate a full-size commit.** Once any order is null, [`js/entry-diff.js:108`](https://github.com/Fliplet/fliplet-widget-data-source/blob/3fcfd27f4a137b3c858261cb320de627a11ae5e0/js/entry-diff.js#L108) builds the baseline by ID alone, ignoring the numbered-first API ordering. With numeric rows whose IDs follow null rows’ IDs, even a no-op drag renumbers most or all rows. Scaled up, a small drag can again produce the 15,000-entry timeout payload.

The dependency-free Node test setup is now coherent, and the author reports 42 passing tests. There are still no GitHub Actions runs; the latest [Codacy check](https://github.com/Fliplet/fliplet-widget-data-source/runs/96755532838) is `action_required` without exposed annotations.

zeryabkhan91:
All three fixed — `e181463`. Each reproduced first.

The common thread you've exposed across these rounds: **an order is only correct if it reads the same to the manager and to everything else.** The manager asks for `id ASC`, the platform default is `id DESC`, so anything resolved by an id tie-break was never really persisted. All three findings are that principle from different angles, and the tests now assert the sequence under **both** id directions.

**Packed insertion produced duplicate orders.** Confirmed exactly as described — the widening loop compared the gap against the new-row count but ignored rows it had already displaced. Inserting into `0/1/2` wrote `NEW=1`, `B=2`, while `C` kept `2`:

```
orders: [0,1,2,2]   B.id=300, C.id=200
manager (id ASC): A,NEW,C,B      consumers (id DESC): A,NEW,B,C
```

Widening now accounts for displaced rows. Same case:

```
orders: [0,1,2,3]   unique: true
manager: A,NEW,B,C  consumers: A,NEW,B,C   agree
```

**Unordered inserts weren't persisted.** You're right, and the `id ASC` request was papering over it rather than fixing it — the position existed only in the manager's view. A data source holding rows with no order is now **numbered once** when a row is added to it, so the position is real. Middle inserts into such a source work as a consequence:

| | manager | consumers | agree |
|---|---|---|---|
| append to unordered | `A,B,C,NEW` | `A,B,C,NEW` | yes |
| middle insert into unordered | `A,MID,B` | `A,MID,B` | yes |

**Mixed sources could still commit in full.** Confirmed: the baseline ignored that rows read numbered-first, so with numbered rows carrying higher ids than unnumbered ones, a **no-op drag over 20 rows wrote 10 of them**. The baseline now matches the read order — that case writes **0**.

Nothing else moved:

| | writes |
|---|---|
| no-change save | 0 |
| delete one / ten | 0 |
| edit one cell | 1 |
| append | 1 |
| drag 10 → 20 | 11 |
| delete + drag | 12 |
| unordered drag | 21 |
| unordered no-op drag | 0 |

**One cost I want to name rather than bury:** numbering an unordered data source is a whole-source commit, so the *first* insert into a large one is expensive — **15,001 rows for a 15,000-row source**. It happens once and every edit after is cheap. The alternative is leaving the position unwritten, which is the defect above. If you'd rather trade correctness for payload there, say so and I'll gate it.

`npm test`: **46 tests, 46 passing**, run for real. Reverting either fix fails them (1 and 4 respectively). Lint still at parity with `production` — 13 pre-existing, none added.

**A broader note.** This is the fourth round, and each has found something real in a different corner of the same model: an integer `order` that is nullable, non-unique, and tie-broken differently by different readers. I've fixed each case, but the model itself is what keeps generating them. Worth a follow-up ticket to make `order` unique and non-null at the API, rather than continuing to defend it from the client side.

zeryabkhan91:
No new review to answer, so I reviewed it myself — and found three more defects. `334963c`.

Four rounds have each turned up the same shape of problem: an order that is right in the manager and wrong to everything else. Hand-written cases kept missing them, so I stopped writing cases and wrote a generator instead. It builds random data sources — dense, sparse, unordered, mixed, **duplicated**, negative — applies random deletes, inserts, drags and edits, then asserts three things:

- the manager shows what the user arranged
- **every other consumer reads the same sequence**
- no two rows end up sharing an order

with pre-existing pathology in the generated source discounted, so only what the save introduces counts. First run: **1,188 of 4,000 scenarios failed.**

**Two insertions in one save could collide.** Placement read each new row's neighbours from *stored* values, so a row displaced by an earlier insertion in the same pass was still read at its old position. Inserting twice into `0/1/2/3` wrote `NEW2` and `B` both at `2` — and which came first depended on their ids, so the two readers disagreed. Placement now reads the positions decided so far.

**An insertion combined with a drag could collide.** A displaced row was placed against its settled neighbours, but the payload then preferred the value the *reorder* had proposed — `R3` and `NEW0` both landed on `2`:

```
intended : R1,R2,NEW0,R3,NEW2,R0,R4
manager  : R1,R2,R3,NEW0,NEW2,R0,R4     <- wrong
orders   : R3=2, NEW0=2
```

There is now one decision per row, and the displacement wins.

**Reordering an unordered source numbered only the prefix.** That was my optimisation from the last round, and it was wrong: the tail kept no order and was separated by id alone, so it came back **reversed** everywhere except the manager. Persisting a reorder now numbers every row. More writes for that case, and the price of the result meaning the same thing to every reader — which is the whole point of the last three rounds.

**Repeated orders are now treated like gaps** — unusable, so the source is renumbered. Two rows sharing an order are separated by id, which is exactly what cannot be relied on.

**4,000 generated scenarios now pass**, and the generator is committed as `test/ordering-properties.test.js` so it runs on every `npm test` rather than being a one-off. Paths that matter are unchanged:

| | writes |
|---|---|
| no-change save | 0 |
| delete one / ten | 0 |
| edit one cell | 1 |
| append | 1 |
| drag 10 → 20 | 11 |
| delete + drag | 12 |

**47 tests passing**, lint at parity with `production`.

I'd still rather this were fixed at the source. Five rounds have all been the same root cause: an integer `order` that is nullable, non-unique, and tie-broken in opposite directions by different readers. The property test now guards the client against it, but the model is what keeps generating these — a follow-up to make `order` unique and non-null at the API would end the category.

Arpanexe:
Re-review of `334963cb4a014690fd2ae645cecb910a3af77a84`: **changes requested**. The previous ordering-collision issues are fixed, but two blockers remain.

### [P1] Large unordered sources still send the entire dataset

`computeHealedPositions()` renumbers every existing row whenever one row is inserted into a source containing null orders ([implementation](https://github.com/Fliplet/fliplet-widget-data-source/blob/334963cb4a014690fd2ae645cecb910a3af77a84/js/entry-diff.js#L183-L192), [call site](https://github.com/Fliplet/fliplet-widget-data-source/blob/334963cb4a014690fd2ae645cecb910a3af77a84/js/entry-diff.js#L352-L361)).

Running the current implementation with 15,000 null-order rows produced:

- **15,001 commit entries** for a single append
- **15,000 commit entries** for one drag

This recreates the large request and likely the timeout that this PR is intended to fix. The property tests generate only 3–10 rows and do not assert a payload-size bound, so they do not catch this regression.

### [P1] The manager now reverses unhealed sources relative to apps

The manager explicitly queries `order ASC, id ASC` ([interface.js](https://github.com/Fliplet/fliplet-widget-data-source/blob/334963cb4a014690fd2ae645cecb910a3af77a84/js/interface.js#L379-L383)), while the platform's default data-source ordering uses `order ASC, id DESC`.

For rows with null or duplicate orders, the manager and app consumers therefore display different sequences. A no-op or data-only save sends no ordering updates, so this disagreement remains indefinitely. Before this query change, the manager used the same default ordering as consumers.

### Checks

[Codacy Static Code Analysis](https://app.codacy.com/gh/Fliplet/fliplet-widget-data-source/pull-requests/276) currently reports `ACTION_REQUIRED`.

zeryabkhan91:
Both blockers fixed — `96c9f76`. You were right on both, and one of them was a regression I introduced two rounds ago.

**The manager no longer asks for its own sort order.** Requesting `id ASC` made it read rows with a null or shared order in one sequence while apps read the reverse, with nothing written down to reconcile them — and, as you say, a no-op save left that in place indefinitely. Reverted: the manager now reads a data source exactly as everything else does, so the two cannot diverge. That was my mistake, and your framing of it — that it was papering over an unpersisted position rather than fixing one — was the right diagnosis.

**Numbering a whole data source is gone.** On a 15,000-row unordered source:

| | before | now |
|---|---|---|
| append one row | 15,001 | **1** |
| drag 10 → 20 | 15,000 | **21** |
| no-op drag | 15,000 | **0** |

Numbered data sources are untouched: no-change 0, delete 0, edit 1, append 1, drag 11, delete + drag 12.

**What that gives up.** A row cannot be positioned among rows that carry no order — a number sorts it ahead of all of them, and numbering them all is exactly the commit you blocked. So on such a data source a new row lands where the shared read order puts it, not where it was dropped. It lands in the **same place for every consumer**, which is the property that actually matters, and it is no worse than before this PR. Making it exact requires `order` to be non-null and unique at the API; it cannot be done from the widget.

**On the test gap you identified** — correct, and fixed. The generator now builds data sources of up to 200 rows rather than 3–10, and asserts a payload bound: a delete, an edit or a reorder must never commit an entire large data source. It also states plainly what is *not* guaranteed, rather than quietly not testing it:

- positioning within a data source that cannot express order (null or shared orders)
- insertion into a fully packed run of integers, which has to shift the rows after it — there is no value to put between two consecutive ones

4,000 generated scenarios pass. 46 tests, lint at parity with `production`.

**Where I think this leaves the PR.** The 502 it was opened for is fixed and has been since the first commit — a delete on a 15,000-row data source went from 3.22 MB and 14.6s to 143 bytes and 0.19s. Every round since has been about the ordering model underneath it: an integer `order` that is nullable, non-unique, and tie-broken in opposite directions by different readers. I've stopped trying to make the widget compensate for that, because each attempt has produced a new defect — including two of my own that you caught. What is here now is bounded, consistent between readers, and honest about what it does not do.

I'd suggest merging this for the timeout fix and raising the `order` column separately. Happy to open that ticket.

Arpanexe:
Re-review of `96c9f76b1e58091b076feee5990c43f8d1c521d6`: **changes requested**. The previous two blockers are bounded/fixed, but the replacement introduces two correctness issues.

### [P1] Appending to an unordered source now reloads at the beginning

When both neighboring orders are null, the new row is intentionally left without an order ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/96c9f76b1e58091b076feee5990c43f8d1c521d6/js/entry-diff.js#L249-L252)).

Because the platform sorts null orders by descending ID, the newly created—and therefore highest-ID—row sorts first. Reproduction: `[id2, id1] + appended` reloads as `[appended, id2, id1]`. This contradicts the PR's append QA scenario and means the “add row” half of PS-1781 succeeds quickly but changes the row's position.

### [P1] Prefix renumbering can collide with untouched negative orders

The incomplete-source path always assigns `0..lastMoved` ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/96c9f76b1e58091b076feee5990c43f8d1c521d6/js/entry-diff.js#L164-L165)), without considering numeric rows beyond that prefix.

With `A:-1, B:0, C:1, D:null`, dragging A below B writes `A=1` while C remains `1`; reload becomes `B,C,A,D` instead of `B,A,C,D`. This mixed-negative state is reachable because inserting above order `0` creates negative orders.

### [P2] The property generator still models the removed `id ASC` manager query

Its initial visual order uses ascending IDs for null/shared orders ([`test/ordering-properties.test.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/96c9f76b1e58091b076feee5990c43f8d1c521d6/test/ordering-properties.test.js#L117-L130)), while the manager now uses the platform's descending-ID default. It also generates `mixed` and `negative` as mutually exclusive source types, which is why the collision above passes the suite.

### Checks

There are still no GitHub Actions runs. [Codacy](https://app.codacy.com/gh/Fliplet/fliplet-widget-data-source/pull-requests/276) remains `ACTION_REQUIRED`.

zeryabkhan91:
Fixed in `63da089`. All three were real; the second one was the more interesting, and chasing it turned up a fourth.

### [P1] Appending to an unordered source reloaded at the beginning

Correct, and the reasoning behind leaving the row unnumbered was wrong. I had it as "there is no gap to sit in", but the row does not need a gap — it needs the rows above it to carry orders, because an unnumbered row always sorts after a numbered one. Leaving it null put it in a pool ordered by descending id, where it holds the highest id, so it read first.

The rows above are now numbered, which is the only thing that can hold the new row where it was put. That is capped at `MAX_ANCHOR_ROWS` (500), because on a data source that carries no order the prefix above an append *is* the data source, and numbering 15,000 rows to hold one of them is the commit this PR exists to remove.

So the trade is explicit rather than hidden:

| unordered source | append lands | writes |
|---|---|---|
| 20 rows | at the end, as arranged | 21 |
| 15,000 rows | first, where every unnumbered row in it reads | 1 |

I could not find a third option. To read after *k* rows, a row needs those *k* rows numbered — that follows from `order ASC NULLS LAST, id DESC`, not from anything in this widget. Any cap therefore buys correctness for small sources at the price of a size-dependent boundary; without one, adding a row to a large unordered source 502s. I picked the boundary that keeps the save working, and it is a named constant with the measurement behind it in the comment. If you would rather have no cap and take the timeout, or a different number, say so and I will change it — it is one line.

### [P1] Prefix renumbering colliding with untouched negative orders

Real, and your repro is exact. `computeReorderedPositions` numbered `0..lastMoved` with no regard for what the rows past `lastMoved` still held, so `A:-1, B:0, C:1, D:null` answered a drag by writing `1` over C. The prefix is now numbered below the lowest order held after it — your case writes `B:-1, A:0` and reloads `B,A,C,D`. There is a named test for it.

### [P2] The property generator modelled the removed `id ASC` query

Also right, and this is the part I should have caught when I reverted that query — I changed the read back and left the generator describing the old one. Fixed: the baseline is built the way the platform sorts, ties on descending id, and `mixed`/`negative` are no longer mutually exclusive (there are now combined shapes plus a random one, so nulls, negatives and duplicates can occur together).

Checked by mutation rather than by assertion: with the collision reintroduced, the old generator reports **0 of 4000** failures and the new one reports **12 of 4000**. Your read of why it passed was correct.

### One you did not raise

Measuring the above turned up that inserting near the top of a *packed* 15,000-row source wrote **14,997** rows — the widen loop walks down looking for slack and a dense sequence has none until the end. Same timeout, and reachable before this branch. The rows above can be moved down instead, so whichever side is fewer rows now gives. That insert is **5** writes.

### Where it stands

| 15,000 rows | numbered | unordered |
|---|---|---|
| no change / no-op drag | 0 | 0 |
| delete a row | 0 + 1 delete | 0 + 1 delete |
| edit a cell | 1 | 1 |
| append a row | 1 | 1 |
| insert after row 3 | 5 | 5 |
| drag row 10 → 20 | 11 | 21 |
| delete + drag together | 11 + 1 delete | 21 + 1 delete |

48 tests and 4,000 generated scenarios pass. Lint I could not run — `npm install` in this repo times out against the registry for me right now, so I have gone over the new code against the config by hand and Codacy will be the real check.

For QA, one expectation changes: adding a row to the end of a data source whose rows carry no order now stays at the end on reload, unless the source is larger than 500 rows, where it reloads at the top. Happy to write that into the plan.

The `order` column being nullable and non-unique is what every round of this has been about, and it is still the case after this. My suggestion stands: merge this for the timeout, raise that separately, and I will open the ticket.

zeryabkhan91:
Fixed in `bf6d384`. The Critical was real and the reproduction was exact — I had reported only the two cheap ends of that curve, which is how it stayed hidden through five rounds.

### Critical #1 — mid-source insert committing half the data source

Confirmed before changing anything, same numbers:

```
insert at 0 → 1   at 3000 → 3001   at 7500 → 7501   at 12000 → 3001   at 15000 → 1
```

The cause is not the algorithm choosing badly — it is that `order` is an **integer** column (`models/dataSourceEntry.js:80`, `integer` in the live schema), so nothing fits between two consecutive values. On a packed `0..n-1` sequence one side has to shift, and a bounded window cannot help: a window of *w* rows spans exactly *w* values, and *w* rows plus one new row need *w+1*. Always one short. `min(above, below)` is a floor, not an implementation choice.

So I took your second option and bounded it. `MAX_ANCHOR_ROWS` is now `MAX_ORDER_WRITES` and governs both expensive shapes — numbering rows above an insertion, and shifting rows around one — with one number and one rule:

| insert into a packed 15,000-row source | writes | lands |
|---|---|---|
| row 0 / 3 / 100 | 1 / 4 / 101 | where you put it |
| row 500 | 501 | where you put it |
| **row 501 → 14,998** | **1** | at the end |
| row 14,999 / 15,000 | 2 / 1 | where you put it |

Past the limit the row keeps its place in the grid for the session and reloads at the end. That is a worse outcome than exact placement and a better one than a seven-second save, and it is now in the PR body under **Limits** rather than left for QA to discover.

The `!hasInserts` exclusion is off for the case you asked for:

```js
var boundApplies = rows.length >= 40 && (usable || (!rowsMoved && !hasInserts));
```

Inserts into a data source that can express order are now inside the bound; the exclusion survives only where holding a row in place genuinely means numbering the rows above it, and that trade is asserted directly instead. Mutation-checked — removing the shorter-side optimisation makes the generator report `commit covered the whole data source (199 of 198)`.

### Warning 1 — column sort plus a drag

Real, and worse than the diff suggests: it is the only remaining path that could still write 15,000 rows. Rather than only fix the diff, the grid now refuses a row move while a sort is applied (`beforeRowMove`), so the user is told instead of having the drag silently dropped. Behind it, `computeCommitPayload` takes `viewMatchesStoredOrder` and reads no position off a sorted grid at all — including for inserts, where the row above a new one is not the row it will reload after. Three tests cover it.

I did not verify in the browser that HoT 4 permits the drag; the guard is correct either way, and the diff-level guard holds even if the grid one never fires.

### Warning 2 — sort-then-save no longer reorders

Documented in the PR body as a behaviour change, with its own QA step. You are right that it is user-visible and right that it should not go back — it is the same write that caused the 502.

### Warning 3 — test seam

Partly closed. The decision that was untestable is now a value: `viewMatchesStoredOrder` is computed in `interface.js` from `table.isColumnSorted()` and asserted in the diff's own specs. What remains untestable without a DOM harness is the flag plumbing itself, so QA step 8 walks the transitions.

### Warning 4 — `applyAndRead()` still on `id ASC`

Correct, and the same defect I had just fixed in the generator. Fixed, along with the docstring. As you predicted, all 56 specs still pass — no assertion was green-but-wrong today.

### Warning 5 — the placement cliff

In the PR body now, with the measured boundary (500 rows → placed, 501 → appended) rather than a description.

### Warning 6 — Codacy

Checked, and it is not this diff. Every PR on this repo based on a `projects/*` branch is `ACTION_REQUIRED` — this one, #272 and #275 — and every PR based on `production` or `master` is `SUCCESS` — #270, #273, #274. It has been that way on this branch since its first commit, before any of the code under review existed. The check exposes no title or summary through the API, so someone with Codacy access needs to open it; I do not think it is waivable-by-argument, but it is also not attributable here.

Lint I still could not run — `npm install` times out against the registry from this machine, including into a clean directory — so the new code was checked against the config by hand, and I renamed the identifiers that shadowed the enclosing scope.

### Suggestions

`hasEntryChanged`'s `positions` parameter is gone; it compares data, and the position decision sits with the caller that has the context for it. Two direct tests added so the export is not dead weight. `widget.json` is back to its original formatting — the deploy diff is one added line. The note about `#275` keeping the originals map scoped to the loaded rows is in the PR body.

56 specs and 4,000 generated scenarios pass. The one thing I have not done is close the gap underneath all of this: `order` being nullable and non-unique is why placement has a price at all, and every round of this review has been a consequence of it. I would still merge this for the timeout and raise that separately.

zeryabkhan91:
Fixed in `e8968ba`. The blocker was real and my guard was worse than no guard.

### Critical #1 — `isColumnSorted()` stuck true after the sort is cleared

Confirmed by transcribing `setSortingColumn` verbatim out of the shipped `handsontable.full.min.js` (0.38.0) and driving the header-click cycle through it:

```
click | sortColumn | sortOrder  | plugin.isSorted() | GridSort.isSortApplied()
    1 |          2 |       true |              true |                    true
    2 |          2 |      false |              true |                    true
    3 |          2 |  undefined |              true |                   false
    4 |          2 |       true |              true |                    true
```

Your reading is exact: `setSortingColumn` clears `sortColumn` only when called with `undefined`, and `sortByColumn` — its only caller — always passes an index, so `isSorted()` means "a header has been clicked", not "a sort is applied". I checked both shipped versions (0.34.5 and 0.38.0 are the only ones in `assets.json`); the definition is identical in each. My comment about Handsontable 4 came from the `afterColumnSort` comment already in the file and was simply wrong about which version this widget loads.

The predicate now reads `hot.sortOrder`, which is what the plugin's own `sort()` tests, with a `typeof` check — you are right that a truthiness test would have failed open on every descending sort, which is the same defect facing the other way.

It moved into `js/grid-sort.js` so a spec can drive it. That is the seam you flagged as Warning 3 last round, and it is exactly where this landed: `it('still writes positions once the sort is cleared')` passed `true` in as a literal, so it proved the module and could never have proved the predicate. The new specs walk a fake instance through the real three-click cycle and assert the descending case explicitly.

Worth stating plainly: this was a guard I added last round that broke two working features for anyone who had touched a column header, and the test I wrote alongside it was shaped so it could not catch that.

### Warning 1 — the refused drag was invisible

Agreed, and the distinction you draw is the right one — the header-row refusal is a defensive guard, this is a new rule the user needs told about. A refused drag now writes into the existing `.entries-message` area under the grid ("Rows cannot be moved while a column is sorted. Clear the sort to reorder rows."), clearing after a few seconds. No modal for a drag.

### Warning 2 — Codacy, and my explanation of it

My paragraph was wrong and I have replaced it rather than softened it. #269 (base `production`) and #271 (base `master`) are both `ACTION_REQUIRED`, so the base-branch correlation does not exist. What the statuses do line up with across the repo is the author:

| PR | base | author | Codacy |
|---|---|---|---|
| #269 | production | hcarneiro | ACTION_REQUIRED |
| #270 | production | hcarneiro | SUCCESS |
| #271 | master | zeryabkhan91 | ACTION_REQUIRED |
| #272 | projects/PR-9 | zeryabkhan91 | ACTION_REQUIRED |
| #273 | production | farhantariq12b | SUCCESS |
| #274 | master | farhantariq12b | SUCCESS |
| #275 | projects/PS-1781 | zeryabkhan91 | ACTION_REQUIRED |
| #276 | projects/PS-1781-order | zeryabkhan91 | ACTION_REQUIRED |

Every PR by this author is red, every PR by `farhantariq12b` is green, on both bases. That points at Codacy account access rather than at any diff, but I am not calling it settled this time — the check exposes no title or summary through the GitHub API, so it needs someone with Codacy access to open it. The PR body now says that instead of the previous claim.

### Warning 3 — placement past the limit is silent at save time

Agreed, and it belongs with the `order`-column work rather than here. I will raise that ticket covering both: making `order` non-null and unique at the API, and telling the user when a row could not be pinned.

### Suggestions

The `applyAndRead()` docstring is corrected — I had rewritten the wrong one of the two last round, so the misleading text was still sitting above the fixed code. The unreachable Handsontable-4 fallback is gone with the predicate it belonged to.

60 specs, 4,000 generated scenarios, all passing. Lint still cannot run here — the registry times out even into a clean directory — so the new module was checked against the config by hand.

zeryabkhan91:
Fixed in `e6ed91e` and `f63b71e`.

### Warning 1 — `GridSort` undeclared

Right, and it would have landed red. `js/spreadsheet.js` now carries `/* global GridSort */`, matching `js/interface.js:1` for `EntryDiff`.

I also got lint running for the first time on this branch, by borrowing the `eslint` in a sibling checkout — so this is measured rather than argued:

```
branch      : 21 problems (0 errors, 21 warnings)   js/ and test/
production  : 21 problems (0 errors, 21 warnings)   same four files
```

Exact parity, and the branch figure includes `entry-diff.js`, `grid-sort.js` and the whole of `test/`, which contribute nothing. Removing the directive again reproduces your finding precisely — `1016:12 error 'GridSort' is not defined no-undef`. Caveat on the rule set: `eslint-config-fliplet` needs `@stylistic/eslint-plugin-js`, which isn't installed anywhere here and can't be fetched, so I ran `no-undef`, `no-redeclare`, `no-loop-func`, `no-unused-vars` and `no-shadow` with the widget's own globals. The stylistic rules are still hand-checked.

### The bigger gap behind that

You could not verify the suite, Arpan could not either, and neither could Codacy — three rounds of "60 specs pass" on my word alone. The repo had no Actions at all, so **`.github/workflows/test.yml`** now runs it on every push and PR. It is green on this head:

```
test  Run the test suite  # tests 60
test  Run the test suite  # pass 60
test  Run the test suite  # fail 0
```

No install step — the suite has no dependencies — so nothing new to keep in step with the lockfile.

### Warning 2 — Codacy

Agreed and unchanged: it needs someone with Codacy access to open the run. The author correlation is in the body as a hypothesis, not a conclusion.

### Suggestions

All three taken.

`showGridNotice` now only clears `.entries-message` if the text is still the one it wrote, so a load error arriving inside the five seconds survives, and the timer is cleared in `destroy()` so a dead grid's notice cannot wipe a live one's. Narrow as you say, but it was a real ordering bug and cheap to close.

`test/grid-sort.test.js` cites `setSortingColumn` in the shipped bundle, names both versions, and says to re-check on a Handsontable upgrade.

Insert-while-sorted losing its position is the same class as placement-past-the-limit, and both go into the `order`-column ticket rather than being papered over here.

### Where Arpan's fourteen findings stand

Re-verified each of his reproductions against this head rather than from memory — his exact inputs, run through the shipped module:

| Round | Finding | Now |
|---|---|---|
| R1 | data-only edit reorders a sparse source | 1 entry, no `order` sent |
| R1 | insert uses a dense visual index | append to `[0,10,20]` reloads `A,B,C,NEW` |
| R1 | delete + drag recreates the payload | 11 updates + 1 delete on 15,000 rows |
| R2 | inserts lose their visual position | middle insert reloads `A,NEW,B,C` |
| R2 | null-order reorder → 15,000 updates | 21 |
| R2 | `package-lock.json` out of step | jest gone; the three devDeps are all in the lockfile |
| R3 | unordered inserts fixed only in the manager | `source.find({})`, no explicit sort — manager reads as apps do |
| R3 | packed insert creates duplicate orders | `A,NEW,B,C` with the id tie-break either way, orders unique |
| R3 | mixed numeric/null → full commit | no-op drag writes 0 |
| R4 | large unordered sources send everything | append to 15,000 null-order rows: 1 entry |
| R4 | manager reverses unhealed sources vs apps | same query as the platform |
| R5 | append to an unordered source reloads first | reloads last (`…,U1,U0,NEW`) |
| R5 | prefix renumbering collides with negatives | `A:-1,B:0,C:1,D:null` drag → `B,A,C,D`, orders unique |
| R5 | property generator models the removed `id ASC` query | generator reads as the platform does; catches the collision in 12 of 4,000, where the old one caught 0 |

10 of the 14 are behavioural and I ran them as a suite — all 10 pass. The other four are structural (lockfile, read order ×2, generator) and are checked above. Nothing of his is outstanding.

I will raise the `order`-column ticket — non-null and unique at the API, plus telling the user when a row could not be pinned — and link it here.

zeryabkhan91:
Both workflow suggestions taken, in `6c0544b`.

**Duplicate runs.** You were right, and the evidence was sitting in the run list — `f63b71e` fired twice, once as `push` and once as `pull_request`. Pushes now cover only `production` and `master`; everything else is covered by its PR. Confirmed on the new head:

```
6c0544b  pull_request  success
f63b71e  pull_request  success
f63b71e  push          success
```

One run where there were two, and no branch loses coverage.

**Token scope.** The job declares `permissions: contents: read` at job level. It checks out the code and runs a suite with no install step, so that is all it needs.

That leaves Codacy as the only red check, and I agree it cannot be closed from here — the check exposes no title or summary through the GitHub API, and the author correlation in the body is a hypothesis rather than a diagnosis. It needs someone with Codacy access to open the run.

On the deferred work — `order` non-null and unique at the API, plus telling the user at save time when a row could not be pinned, whether past `MAX_ORDER_WRITES` or inserted while a sort is applied — I agree it belongs in its own ticket and not in this PR. I will raise it and link it here once I have the go-ahead on where it should sit.

Thanks for the three rounds. Two of the defects you found were guards I had added myself while answering earlier rounds, and in both cases the test I wrote alongside was shaped so it could not have caught them — that is the part I have tried to fix structurally rather than case by case, which is what the property generator, the extracted predicate and now CI are for.

Arpanexe:
Re-review of `6c0544bd4dd53d3a3f39e514160b2549105211a8`: **changes requested**. The earlier findings are fixed, but one P1 remains.

### [P1] Multiple inserts do not compose on unordered sources

Placement state for newly inserted rows is not reused by subsequent insertion runs; `anchorRowsAbove()` aborts when it encounters any earlier new row ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/6c0544bd4dd53d3a3f39e514160b2549105211a8/js/entry-diff.js#L275-L283)).

Reproduction with three null-order rows:

```
intended: A, N1, B, N2, C
reloads:  A, N1, N2, B, C
```

The second row remains unordered and its newer ID moves it ahead of B.

Two consecutive rows inserted at the top also reverse because the no-anchor branch leaves both unordered ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/6c0544bd4dd53d3a3f39e514160b2549105211a8/js/entry-diff.js#L439-L447)):

```
intended: N1, N2, A, B, C
reloads:  N2, N1, A, B, C
```

Both cases use only three existing rows—well below the 500-write cap—so they should preserve placement.

The property generator misses this because it only compares reloaded order with the intended arrangement when every original row already has a numeric order ([`test/ordering-properties.test.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/6c0544bd4dd53d3a3f39e514160b2549105211a8/test/ordering-properties.test.js#L202-L208)).

### Checks

The new [GitHub Actions test job](https://github.com/Fliplet/fliplet-widget-data-source/actions/runs/32718968331/job/97406122255) passes. [Codacy](https://app.codacy.com/gh/Fliplet/fliplet-widget-data-source/pull-requests/276) remains `ACTION_REQUIRED`.

zeryabkhan91:
Fixed in `3e8c3e2`. Both reproductions confirmed exactly as written before I changed anything.

### [P1] Multiple inserts not composing

Two faces of one mistake: each run of new rows was placed as if it were the only one in the save.

A run read its neighbours through a helper that only knew about stored rows, so a row added earlier in the same pass read as unnumbered — and `anchorRowsAbove` then refused outright to number a prefix containing one. Your first case, verbatim:

```
intended  A, N1, B, N2, C
reloaded  A, N1, N2, B, C     orders  A:0  N1:1  B:null  N2:null  C:null
```

The top-of-source case is the same problem from the other side: with nothing numbered above or below, both rows were left unnumbered, and unnumbered rows read newest first.

```
intended  N1, N2, C, B, A
reloaded  N2, N1, C, B, A     orders  all null
```

Placement now reads and writes both kinds of row through one pair of position-keyed helpers, so a row added earlier in the pass is a neighbour like any other. A run at the top is numbered when there are several of it; a single row there is still left alone, since it holds the newest id and therefore already reads first — correct for nothing.

You were right that these are cheap cases. Both are three-row data sources, nowhere near the write limit. The costs are unchanged: 1/4/101/501 on a packed 15,000-row source then 1 past the limit, and I re-ran your earlier fourteen findings — all still hold.

### The generator, which is the part worth dwelling on

Your diagnosis was right and I had it half-fixed. Two things blinded it:

**The gate.** Comparing the reload against the intended arrangement only when every stored row already held a number excluded exactly the sources the defect needed. That is now gated on ambiguity instead — a source with orders shared between rows has no defined intended sequence, but one that simply carries no order does, and placement is owed on it whenever it is small enough to renumber.

**The generator itself.** It inserted strictly one row at a time, so a run of two at the top was a 3-in-4000 accident. It now inserts runs of two or three about a third of the time.

I checked by mutation rather than by assertion, because widening the gate alone looked like it worked and did not:

| defect reintroduced | old suite | gate widened only | gate + runs |
| --- | --- | --- | --- |
| `anchorRowsAbove` bails on a new row above | 0 / 4000 | 47 / 4000 | 47 / 4000 |
| a run at the top left unnumbered | 0 / 4000 | **0 / 4000** | 11 / 4000 |

The middle column is the reason I am reporting this in that much detail — had I stopped at the gate change I would have told you the generator now covers this, and for the second case it would not have. Five named tests pin both reproductions and three more shapes besides, so neither depends on the random stream.

65 specs, 4,000 generated scenarios, lint 21 warnings / 0 errors — the same 21 production reports. Actions green on the head.

Codacy is still `ACTION_REQUIRED`, unchanged and still not readable from here.

zeryabkhan91:
Rather than wait for the next round, I swept the scenario space for this class of defect. `d741887`.

**What I ran:** ten order shapes (packed, sparse, unordered, mixed, negative, mixed-negative, head-null, tail-null, duplicated, all-zero) × three id layouts (sequential, reversed, shuffled) × five sizes × every delete, edit, insert of one/two/three rows, drag and combination at five positions in each — **3,660 cases**, plus **220** more at 100 / 501 / 1,200 / 3,000 rows where the write limit bites. Five invariants each: no duplicate orders introduced, the row set matches what was left on screen, the reload matches the arrangement, the save introduces no dependence on the id tie-break, and — the one nothing had checked before — **a save settles**.

**One defect.** A single row added at the top of a data source carrying no order was left unnumbered. My reasoning was that it holds the newest id and unnumbered rows read newest first, so it already reads first for nothing. True of the read order today, and it stores nothing: the position rests on the tie-break rather than on the data source, and a source that was unambiguous before the save is ambiguous after it. Numbering it costs no extra row — the row is in the commit either way, this only fills in its `order` — so a run at the top is now numbered whatever its length.

Everything else held, including the cost bounds, your fourteen earlier findings, and both of yesterday's reproductions.

### The part that concerns me more than the defect

Two gaps let it through, and the second is the one worth recording.

**The suite never re-saved.** Nothing anywhere checked that a save *settles*. A first save could look completely correct — right sequence, no duplicates, small payload — and still leave the manager and the stored order disagreeing, so every save afterwards rewrites rows nobody edited. That is the failure mode of the original bug, arrived at from a different direction, and it was untested. The generator now applies its own payload, saves again untouched, and requires both that and a following no-op drag to write nothing.

**The generator only ever built sources of three rows or more**, and this class needs one or two. It now builds them, and asks whether the save *introduced* a tie-break dependence rather than only checking sources that were numbered to begin with.

Mutation results — reintroduce each defect from this round and count what the suite catches:

| defect reintroduced | before | now |
| --- | --- | --- |
| single top insert left unnumbered | 0 / 4000 | 5 / 4000 |
| `anchorRowsAbove` bails on a new row above | 0 / 4000 | 43 / 4000 |
| a run at the top left unnumbered | 0 / 4000 | 29 / 4000 |
| whole source renumbered on a drag | not caught by the settle check | caught by it alone |

66 specs, 4,000 generated scenarios, 3,880 enumerated ones, lint 21 warnings / 0 errors — the same 21 production reports. Actions green.

Codacy unchanged and still unreadable from here.

Arpanexe:
Re-review of `d7418873057efbb01e83a79f7e6bd19be3ec4459`: **changes requested**. The previous multiple-insert finding is fixed, but two cap-boundary failures remain.

### [P1] An adjacent drag near the bottom of an unordered source still recreates the timeout payload

For incomplete order data, `lastMoved` is converted into a rewrite of every row from index `0` through that point ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/d7418873057efbb01e83a79f7e6bd19be3ec4459/js/entry-diff.js#L177-L210)).

On 15,000 null-order rows, moving row `14900 → 14901` produces **14,902 updates**. `MAX_ORDER_WRITES` only protects insertion placement; reordering bypasses it. An ordinary one-row drag can therefore still produce the same multi-megabyte commit and 502.

### [P1] Combining a large unordered reorder with multiple inserts introduces duplicate orders

Reproduction on 1,200 null-order rows:

1. Insert a row at index 300.
2. Move the row at index 1100 up one position.
3. Insert another row at index 511.
4. Save.

The payload contains **1,100 existing-row updates** and assigns both `R689` and `R690` order `510`; the second inserted row also reloads outside its intended position.

Placement updates override reorder decisions ([`js/entry-diff.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/d7418873057efbb01e83a79f7e6bd19be3ec4459/js/entry-diff.js#L618-L628)) without reconciling the untouched reorder assignments.

The committed property generator cannot exercise this boundary because its largest generated source is 219 rows ([`test/ordering-properties.test.js`](https://github.com/Fliplet/fliplet-widget-data-source/blob/d7418873057efbb01e83a79f7e6bd19be3ec4459/test/ordering-properties.test.js#L118-L123)).

### Checks

The latest [GitHub Actions test job](https://github.com/Fliplet/fliplet-widget-data-source/actions/runs/32723661349/job/97420173709) passes. [Codacy](https://app.codacy.com/gh/Fliplet/fliplet-widget-data-source/pull-requests/276) remains `ACTION_REQUIRED`.

zeryabkhan91:
Closing out the lint caveat I have been carrying for three rounds, in `ce18234`.

`eslint-config-fliplet` needs `@stylistic/eslint-plugin-js` for exactly **one** of its 107 rules — `comma-spacing`. That single missing plugin was why I could not run it at all, and why I kept reporting a five-rule substitute. Pointing the other 106 at an eslint installed elsewhere on this machine gets the real rule set running.

It found one problem, and it was mine: `no-nested-ternary`, in the source-size expression I added in the previous commit. Now a named function, which reads better than the ternary did and leaves the random stream untouched.

With that fixed:

```
branch      13 problems (7 errors, 6 warnings)
production  13 problems (7 errors, 6 warnings)
```

Identical rule for rule and file for file — `operator-linebreak` ×3, `space-in-parens` ×2, `no-unused-vars` ×2, `no-console` ×6 — differing only in line numbers, since the branch adds lines to those two files. `js/entry-diff.js` and `js/grid-sort.js` report nothing under the full set.

Worth noting the number: 13 is what I claimed as the production baseline in the first round, from a run I could not reproduce afterwards. It checks out.

The PR body no longer carries the "stylistic rules checked by hand" caveat. 66 specs, 4,000 generated scenarios, 3,880 enumerated ones, Actions green. Codacy unchanged.

zeryabkhan91:
Two commits since `d741887` — `29430f8` and `9da7a94`. Both P1s from the last review are fixed, a third defect nobody had reported turned up while fixing them, and a later review of `29430f8` raised four more, all four of which reproduced. Everything below was run against the module rather than reasoned about.

## The two P1s

**[P1] An adjacent drag near the bottom of an unordered source.** Reproduced at 14,902 updates, exactly as reported.

The cause was that the bound only ever governed insertion. `MAX_ORDER_WRITES` now governs a drag too — but counted by **what it writes, not the span it covers**, which matters more than it sounds. Renumbering a prefix that is already numbered `0..k` hands most rows back the value they were holding and commits only the few that moved. Bounding on the span instead would have refused those drags outright, which is a regression I nearly shipped: on a data source numbered with a short unnumbered tail, a deep drag costs **2 rows** and must keep working. There is a spec on exactly that shape now.

| 15,000 rows, one adjacent swap | before | now |
|---|---|---|
| unordered, at index 14,900 | 14,902 | **0 — declined** |
| unordered, at index 10 | 21 | 21 |
| numbered, any depth | 2 | 2 |
| numbered + unnumbered tail, at index 14,000 | 2 | 2 |

**[P1] Large unordered reorder plus multiple inserts.** Reproduced exactly — 1,100 updates and `R690`/`R691` both on order 510. Now 0 updates, both new rows unnumbered, no duplicate.

Your note that the property generator could not exercise this was the most useful part of the review. Its largest source was 219 rows, so no invariant it checked could ever reach the 500-row boundary where all of this behaviour turns on. It now builds sources past the limit, and it catches the defect: 1 of 4,000 scenarios fails on the old module, 0 on the new one.

## A third one, not reported

Fixing those exposed a defect on the shape you would least want it on — **a fully numbered data source, two rows added in one save, no drag at all**:

> 1,200 rows ordered `0..1199`. Add a row at index 505 and another at 512, then save.
> The payload pushes `P505..P510` down to make room, stops at `P510` and writes order **511** — which `P511` is holding and the save never touched. The second new row gets no order and reloads at the end.

The widening walk stops at the first row with no settled order, and a row inserted later in the same save reads exactly like the end of the data source. So the run was written into space the rows past it still held. New rows met mid-walk are carried along now, and a run already placed that way is not placed twice. A packed `0..n-1` sequence is what the *old* save left behind, so this was reachable on any data source this manager had saved before.

## What the customer's data source actually looks like

Read from the prod EU replica, since the whole scoping question rested on it — `Registration`, ds `326119`, 14,761 live rows:

- **14,748** uniquely numbered `0..14,748`
- **13** carrying no order at all — every one created between 18 and 24 August, all with ids above every numbered row

That is the API's behaviour visible in the data: `order` is a plain integer with no default and nothing in the entry-creation path sets one, so every row added since the last successful manager save has none. The last save numbered up to 17 August; the 502 was re-reported on the 19th.

It also corrects the first P1's scope, so worth stating plainly: **it was never live for this customer.** The prefix renumbering writes thousands of positions but only rows whose value actually *changes* reach the payload, so on their shape a deep adjacent drag commits 2 rows even on the unfixed code. What *was* live for them is a long drag:

| on ds 326119 | `ce18234` | now |
|---|---|---|
| delete 5 rows from the middle | 0 updates + 5 deletes | same |
| add a row at the end | 14 rows | same |
| drag 14,000 → 14,001 | 2 rows | same |
| **drag row 0 → 14,700** | **14,701 rows, 1.47 MB** | **declined** |

## The four findings from the review of `29430f8`

**1. A row added while a column is sorted lost its position and said nothing — correct, and the table reproduces line for line.** The fallback that skips the placement pass hard-coded `unplaced: 0`, so `declined.rows` was structurally unreachable on both paths that skip it. The notice was wired to every decline except the two that skip placement outright. Skipped rows are counted now, and the payload names the reason, so the message can say the one useful thing about a sort — that clearing it puts placement back.

**2. The refused-drag message did not mention the new rows — correct.** 600 unordered rows, drag 590→580 plus an append: the row the user had just added reloaded at index 0 of 601 while the message spoke only about row order. It now says the rows added moved with it. The docstring on `MAX_ORDER_WRITES` had the same fault — it said a declined row "reloads at the end", when on an unordered source it reloads at the **top**. Corrected.

**3. The test seam — taken.** `commitNotice` is `js/commit-notice.js` now, extracted the same way `GridSort` was, with eight specs. The prediction was exact: it would have caught finding 1, because the sorted branch was the one no caller exercised.

**4. The `!reorder.refused` guard — right about the test, and I can now settle the mechanism.** It was unpinned; removing it left 74/74 green, which is on me. It is not dead code, though. Re-running the 29,358-case sweep against a guard-removed copy brings the collision back — 20 hits — and it reduces to:

> 502 rows numbered `0, 3, 6…` plus 2 unnumbered. Insert at index 501, then swap rows 500 and 499.
> Without the guard the new row takes order **1500**, which `M500` is still holding.

It needs three things at once — a numbered head long enough to exceed the write limit, an unnumbered tail to make the reorder refusable, and the new row *inside the span the drag disturbed* — at which point the visual neighbours are in an order the data source does not hold and the row below reads as the row above. One insert is enough. A spec drives exactly that case, and all three new guards are mutation-checked: removing the refusal, the count, or the reason each turns the suite red.

## Two things QA needs told, both now in the PR body

1. **A drag past the limit is not persisted** — only on a data source that carries no order. The row returns to its stored position on reload, and the grid says so.
2. **A row added while a column is sorted is not positioned** — it reloads at the end of a numbered source, or at the *top* of an unordered one. The grid says so, and says that clearing the sort puts placement back.

The notice fires on every row added under a sort, including an append that would have landed correctly anyway. That is deliberate — under a sort no placement was derived from the grid at all, so the statement is true in every case — but it is listed as narrowable if it reads as noisy in QA.

## State

86 specs and 4,000 generated scenarios pass, green on Actions. Beyond the suite: 29,358 insert/drag scenarios across unordered, packed and mixed sources — 856 duplicate-order cases on the old module, **0** now; and 1,500 punishing saves of 600–3,200 rows with deletes, insert runs and drags together — no duplicates, nothing commits a whole data source, largest commit 2,878 entries (~0.6 MB).

The PR body's **Limits** section is rewritten around what actually drives cost, and carries a **Known limits this does not close** section: a long drag on a *fully* numbered source is still unbounded (~15,000 rows, ~3 MB — it needs zero unnumbered rows, so any source with an addition since its last save takes the bounded path), the limit is per path rather than per save, and `order` itself is still nullable and non-unique. That last one is the cause underneath every round of this review, and I will link the ticket here.

Lint is unchanged from `production` — same problems, same files, rule for rule; `js/entry-diff.js`, `js/grid-sort.js` and `js/commit-notice.js` report none. Codacy is still `ACTION_REQUIRED`, as it has been since this branch's first commit and on every PR by this author, and still needs someone with Codacy access to open it.

Arpanexe:
Review of head `9da7a94`:

### [P2] Guard generated orders against signed INTEGER overflow

The arithmetic around [`base`/`next`](https://github.com/Fliplet/fliplet-widget-data-source/blob/9da7a94ccd72e222ff8235e12bcdbe98fad32656/js/entry-diff.js#L568-L587) is unbounded, while `dataSourceEntry.order` is stored as a Sequelize `INTEGER`.

Reproductions against the current `EntryDiff.computeCommitPayload`:

- Append after order `2147483647` → the new row is sent with order `2147483648`.
- Insert before order `-2147483648` → the new row is sent with order `-2147483649`.
- Reorder rows sharing order `-2147483648` on an incomplete source → updates are sent with orders as low as `-2147483650`.

These are outside the signed 32-bit range and will make an otherwise valid save fail at the API/database. Please validate generated orders against `[-2147483648, 2147483647]` and either rebalance within range or decline the position without sending an invalid order. Boundary tests should cover insertion at both extremes and the incomplete-source reorder path.

The previous two P1 findings are fixed in my reproductions. GitHub Tests pass; Codacy remains failing/action-required.

zeryabkhan91:
Fixed in `752c905`. All three reproduce, and the consequence is worse than the finding says — so thank you for it.

## Confirmed, including against the live schema

`order` really is `integer`, precision 32, checked in `information_schema` on the production reader rather than read off the Sequelize model. So a value past either end is not a row in the wrong place: Postgres rejects the write and **the whole save fails**, which is the outcome this PR exists to remove. Right to raise it.

All three reproductions land:

| reproduction | before |
|---|---|
| append after `2147483647` | `NEW = 2147483648` |
| insert before `-2147483648` | `NEW = -2147483649` |
| three rows sharing `-2147483648`, swap the top two | `B = -2147483650`, `C = -2147483649` |

One note in case it saves you time later: the third only reproduces with the rows in **read order**. Rows sharing an order are read by descending id, and handed to the module in id-ascending order the same swap looks like a different move and writes nothing at all. That cost me a wrong "could not reproduce" for ten minutes before I spotted it.

## Five paths, not three

The three reproductions reach three of them; auditing the rest found two more that generate numbers the same way — `pullDownRowsAbove`, which pulls the rows above an insertion down below a ceiling, and `numberRunAtTop`. All five now check that the span they would write fits the column, and decline the position when it does not.

Declining is what these paths already do with a position that costs more than the write limit allows, so this needed no new concept and no new user-facing behaviour: the notice added in `9da7a94` reports it for free.

Measured over 6,000 saves against data sources parked at both ends of the column — packed at the top, packed at the bottom, spread across the whole range, duplicated extremes, and extremes mixed with unnumbered rows:

| | orders outside `int4` |
|---|---|
| `9da7a94` | **3,065** |
| `752c905` | **0** |

873 of those saves now tell the user what they declined.

## On the severity, with evidence

Agreed it is a P2, and here is the reason rather than an assertion. Across every region's `dataSourceEntries`:

| region | min `order` | max `order` | rows beyond ±1,000,000,000 |
|---|---|---|---|
| EU | 0 | 100,000 | 0 |
| US | 0 | 88,605 | 0 |
| CA | 0 | 2,420 | 0 |

Four orders of magnitude of headroom, and nothing anywhere near the ends. Reaching this needs a client writing its own `order` values through the API — the manager cannot produce it, before this change or after. Worth fixing, not worth holding a 502 fix for.

## Tests, as asked

Boundary specs cover insertion at both extremes and the incomplete-source reorder. The property generator also builds the extreme shapes now and asserts the invariant directly — that no order a save writes falls outside the column.

Two things about that I would rather flag than have found:

**The generator's placement assertion had to be widened, and that is a real loosening.** Placement cannot be owed on a data source with no headroom left — there is no number between `2147483647` and "after it" — and the rows that get declined carry no order, so they also tie with each other. Adding the extreme shapes without that gate turned 100 of 4,000 scenarios red for behaving correctly. Both assertions are now gated on the source having room in the column, and the gate is commented where it happens.

**Four of the five guards are pinned; the fifth is not, and says so in place.** Disabling each in turn: the reorder prefix, `anchorRowsAbove`, `pullDownRowsAbove` and the main placement each turn the suite red. `numberRunAtTop` does not, and I could not construct a state that reaches it — a numbered row always reads before an unnumbered one, so the branch it protects cannot see a numbered row below an unnumbered one today. Rather than write a spec that does not exercise it, the comment states that it is defensive and why. You were right about an unpinned guard last round; this one is unpinned by nature, so it is declared.

## State

90 specs and 4,000 generated scenarios pass, green on Actions. The rest of the battery is unchanged: 29,358 insert/drag scenarios with no duplicate orders, 1,500 large multi-op saves with nothing committing a whole data source, and the customer's own data source behaving exactly as before.

Thanks for confirming the two P1s — and for the generator criticism two rounds back, which is what made these boundary shapes possible to add at all.

Arpanexe (approved):
LGTM

